Issue #11, opened 2015-09-05
Building in OSX


rperrot:
Hi, 

I'm trying to build openMVS on OS X (10.10) but I have some issues.

First I installed dependencies using macports : 

```
sudo port install opencv boost cgal ceres-solver eigen3
```

I got vcv_lib and openMVG using the instruction on the build page. I build and installed the openMVG library with success. 

I configured the openMVS project using the provided command. But when I did make I had the following issue : 
```
In file included from /Users/rperrot/Prog/openMVS/build/libs/Common/cotire/Common_CXX_prefix.hxx:4:
In file included from /Users/rperrot/Prog/openMVS/build/libs/Common/cotire/Common_CXX_prefix.cxx:4:
In file included from /Users/rperrot/Prog/openMVS/src/libs/Common/Common.h:176:
/Users/rperrot/Prog/openMVS/src/libs/Common/Types.h:321:11: fatal error: 'endian.h' file not found
# include <endian.h>
          ^
1 warning and 1 error generated.
```

This is because on OS X (using clang) endian.h is not provided as is, it must be included by <machine/endian.h> instead. I created a cmake rule : 

```
if( APPLE )
    LIST( APPEND OpenMVS_DEFINITIONS -D_APPLE_ )
ENDIF( APPLE )
```

and in lib/Common/Types.h I did (line 315 and following) : 

```
#if defined(_MSC_VER)
# define __LITTLE_ENDIAN 0
# define __BIG_ENDIAN 1
# define __PDP_ENDIAN 2
# define __BYTE_ORDER __LITTLE_ENDIAN
#elif defined(__GNUC__)
#if defined(_APPLE_)
#include <machine/endian.h>
#else
#include <endian.h>
#endif
#endif
```

That solve this first issue. But after that I got lot's of other errors : 

``` 
/Users/rperrot/Prog/openMVS/src/libs/Common/Thread.h:149:52: error: use of undeclared identifier 'pthread_setschedprio'; did you mean 'pthread_setspecific'?
        inline void setThreadPriority(Priority p) const { pthread_setschedprio(threadHandle, convertPriority(p)); }
                                                          ^~~~~~~~~~~~~~~~~~~~
                                                          pthread_setspecific
/usr/include/pthread.h:456:5: note: 'pthread_setspecific' declared here
int pthread_setspecific(pthread_key_t , const void *);
    ^
[...]

PR #15, opened 2015-10-02
Fixes for OSX and C++11

Meekohi:
These changes combined with these build commands (https://gist.github.com/Meekohi/c802d381076d02e76ba8) work for me on OSX with most recent libraries installed.


libs/IO/ImageTIFF.cpp
```diff
@@ -10,8 +10,13 @@
 #ifdef _IMAGE_TIFF
 #include "ImageTIFF.h"
 
+// This corrects a conflict where both libtiff and opencv3 try to redefine uint64 and int64
 extern "C" {
-#include <tiffio.h>
```
Meekohi: The only changes other than these lines in this file is to rename all instances of `int64` to `int64_tiff` so there is no name collision with openCV.

libs/MVS/SceneRefine.cpp
```diff
@@ -723,7 +723,7 @@ void MeshRefine::ImageMeshWarp(
 {
 	ASSERT(!imageA.empty());
 	typedef Sampler::Linear<float> Sampler;
-	const Sampler sampler;
```
Meekohi: In C++11 you cannot use the default initializer with a const object.

libs/MVS/SceneTexture.cpp
```diff
@@ -1677,7 +1677,8 @@ void MeshTexture::PoissonBlending(const Image32F3& src, Image32F3& dst, const Im
 			const MatIdx idx(indices(i));
 			ASSERT(idx != -1);
 			coeffA.AddConstruct(idx, idx, 1.f);
-			coeffB[idx] = Color(dst(i));
+			Color c = dst(i);
```
Meekohi: Make it unambiguous as to whether the move or copy implicit type-cast is being used (C++11 won't let you leave it ambiguous).

Meekohi:
Hopefully fixes Issue #11 (needs to be tested) without breaking anything else.

cdcseacave:
Thx for the patch, I'll add it soon.

libs/MVS/SceneRefine.cpp
cdcseacave: it was already fixed (differently) in the last commit

libs/MVS/SceneTexture.cpp
cdcseacave: There should be a nicer solution to this. Could you please try something different to fix it (without creating a temporary variable)?

libs/IO/ImageTIFF.cpp
cdcseacave: I like the trick with the defines. However, please do not modify my code just because of a conflict generated by bad libs. If it is not compiling just with the define trick, please fix it by using the same tiff library.

libs/MVS/SceneTexture.cpp
rperrot: You can just add a cast to a reference (lvalue) to force using copy ctr :

```
coeffB[idx] = Color( (Color&)dst(i));
```

This compiles fine on OSX using clang.

cdcseacave:
build on mac solved

PR #22, opened 2015-10-20
Irreversible file overriding (data corruption & loss)

F0x06:
Changed undistorded images output folder to "images_undistorded" instead of "images", because in many cases you will place source images in "images" folder, and this will result an irreversible data loss, due to same path for source and destination images during copy.

Best regards,
F0x.

### DIFF
--- a/apps/InterfaceOpenMVG/InterfaceOpenMVG2.cpp
+++ b/apps/InterfaceOpenMVG/InterfaceOpenMVG2.cpp
@@ -56,7 +56,7 @@ bool bOpenMVS2PLY; // conversion direction
 bool bNormalizeIntrinsics;
 String strInputFileName;
 String strOutputFileName;
-String strOutputFolder;
+String strOutputImageFolder;
 unsigned nArchiveType;
 int nProcessPriority;
 unsigned nMaxThreads;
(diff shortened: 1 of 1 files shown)


pmoulon:
Dear @F0x06 please note that it can be the same with any directory you have chosen ;-)
An user can have a images_undistorted directory as input...

F0x06:
Dear @pmoulon,

Yes but less chances than images to cause conflicts i think, and this make more sense to save image in "images_undistorded" by the way, its a suggestion, another folder name will also works.

pmoulon:
perhaps the safer solution could be to check if the output directory exists... and if it exists show a message and stop.

Another option could be to add a command line option to choose the output image directory name. So the user can do whatever he wants. (with a default name undistorted_images)

F0x06:
Good idea.

F0x06:
@pmoulon, @cdcseacave 
--output-folder option added

pmoulon:
I don't know if you can change the option to output-image-folder (the actual one can be confusing) we can think that all data will be exported in this dir...

F0x06:
@pmoulon Good point, will do that.

Issue #21, opened 2015-10-20
View filtering


rjanvier:
see OpenMVS/OpenMVS#6
I can implement it on my own, but are you agree with this idea ?
R.

cdcseacave:
Fine with me. Please implement it as an other ececutable in Apps, entirely self contained (i dont think u have to touch any existing code). In order to discard the use of an image, simply set poseID to NO_ID. Also, please use the same coding style used in the rest of the code.

Thank you.

ORNis:
thank you,
I will try to do my best.

cdcseacave:
Closing as stale. The OpenMVS codebase has undergone extensive refactoring and modernisation on the \develop\ branch since this issue was opened. If you believe this is still relevant, please test against the latest \develop\ branch and open a new issue with reproduction steps.

PR #23, opened 2015-10-20
FilterViews

rjanvier:


ORNis:
OpenMVS/openMVS#6
#21

cdcseacave:
Looking good, but i think you need to filter all points too, and remove the
views   that u disabled (and remove the entire point if no more views
available).

ORNis:
Ok thanks a lot,
I will look at this at the end of the week.

rjanvier:
I close this old one for now.

Issue #46, opened 2016-01-26
Building fails: invalid path for Ubuntu in FINDVCG.cmake in Ubuntu 15.10 


mad-de:
Building fails with:

CMake Error at build/Utils.cmake:214 (message):
  VCG required, but not found: Please specify VCG directory using VCG_ROOT
  env.  variable
Call Stack (most recent call first):
  build/Modules/FindVCG.cmake:22 (package_report_not_found)
  libs/MVS/CMakeLists.txt:9 (FIND_PACKAGE)

been trying around a little bit, but haven't been able to fix this one yet. In the current Building.md VCG is installed with: 

svn checkout svn://svn.code.sf.net/p/vcg/code/trunk/vcglib vcglib
So the paths given (    PATHS "$ENV{PROGRAMFILES}" "$ENV{PROGRAMW6432}" "/usr" "/usr/local" "/usr/share" "/usr/local/share"
    PATH_SUFFIXES "vcg" "include") 

shouldn't work anyway unless i'm missing something.

Quite busy right now and in the next days. If you don't have time, i'll try fixing it in the next days, but that may take a while until I'm able to look into it...

cdcseacave:
please use the steps listed in the wiki:
https://github.com/cdcseacave/openMVS/wiki/Building

mad-de:
Did so. The build.md installs vcg as ~/vcglib/vcg (in the home folder). That folder however does not get called by "$ENV{PROGRAMW6432}" "/usr" "/usr/local" "/usr/share" "/usr/local/share"
PATH_SUFFIXES "vcg" "include") in the Buildvcg file.

cdcseacave:
It does not, and there is no need. Set the path to it during CMAKE setup, as showed in the wiki.

mad-de:
alright - I made a PR to set the path for ubuntu to ~ instead of $main_path in the building instructions, which solved the issues I had. Please check it out.

cdcseacave:
please see comment at #47

dxqxiaoqiang:
Did you have solved your question ? I meet with  the same problem as you .

PR #47, opened 2016-01-27
Change $main_path/ to ~/ in Build instructions for Ubuntu

mad-de:
Fixes https://github.com/cdcseacave/openMVS/issues/46


cdcseacave:
main_path should be set to the right path, and not hard code a custom path

PR #51, opened 2016-02-02
add "cd .." to openMVG install

mad-de:
as openMVG is optional, we might need to cd back to the home folder to build openMVG.

Otherwise folder structure is:
~/openMVG
~/openMVG_build
~/openMVG_build/openMVS
~/openMVG_build/openMVS_build


pyp22:
correct. Observed that too.

PR #103, opened 2016-07-11
Added CUDA CMake linker flags to fix errors during linker stage

mdaiter:
Added two extra flags that fix CUDA linker errors during linking libMVS.


cdcseacave:
Thanks for this fix, however I find it more like a hack. Could you please try to solve it by making sure ${CUDA_LIBRARIES} that is already included contains the missing libraries? I suspect there is a problem with FIND_PACKAGE(). Maybe it needs additional params?

cdcseacave:
And failing the continuous integration test proves my point.

mdaiter:
@cdcseacave Just updated linker flags to be integrated only if CUDA was found. Is including linker flags in a string like that hacky, or is that best practice for CMake?

cdcseacave:
This is indeed still a hack (for many reasons, one is that it will not work on windows or if the path to the libs are not added to the env path)! Please fix the find_pakage to correctly fill the missing libs to ${CUDA_LIBRARIES} variable as described in the first comment.

cdcseacave:
Thx @mdaiter, however all those extra libraries are not needed for OpenMVS, as they are CERES dependencies (so linked indirectly). Please use the build process described in the documentation and all works fines, as demonstrated also by the continuous integration.

mdaiter:
@cdcseacave Oh, didn't mean to push haha. I was on NixOS, and for some reason CERES on NixOS isn't compiled with (or is statically linked to) GLOG, so there were manual files that had needed to be included.

bartoszek:
@cdcseacave On Linux cmake is set up to build runtime api applications, for which -lcuda is not required. The application you are building use driver API (things like cuInit are part of the driver API) and your cmake should add -lcuda to the link specification. There is nothing wrong with find_package and there is anything you could do beside what @mdaiter proposed. To keep compatibility with windows wrap the assignment in if statement.
```
if (UNIX)
  SET(CUDA_LIBRARIES "${CUDA_LIBRARIES};-lcuda") 
endif (UNIX)`
```

cdcseacave:
fixed in 54766b9

Issue #109, opened 2016-07-29
Is PointProjectP division applied twice?


paulinus:
This is more a question than an issue.

### Specifications like the version of the project, operating system, and hardware

Master branch on a mac. Both using clang and gcc.

### Steps to reproduce the problem

The camera method [`PointProjectP`](https://github.com/cdcseacave/openMVS/blob/79bacacff692ffd9b8631dcf729f6f8a6c76e8be/libs/MVS/Camera.h#L255-L259) returns a `TPoint2`. The point is build from a `TPoint3` and then divided by the last coordinate, z:
```
return ((const TPoint2<TYPE>&)q)*INVERT(q.z);
```

When debugging the code, i'm seeing that the [following constructor](https://github.com/cdcseacave/openMVS/blob/79bacacff692ffd9b8631dcf729f6f8a6c76e8be/libs/Common/Types.h#L1295) is called to build the `TPoint2`:
```
explicit inline TPoint2(const cv::Point3_<TYPE>& pt) : Base(pt.x/pt.z,pt.y/pt.z) {}
```
This constructor is also doing the division by the last coordinate of the `TPoint3`. So projections end up being divided by z twice.

The result on my tests is that the colors of the pointclouds totally misaligned. Curiously enough, the point cloud it self looks ok, so probably the functions building the pointcloud are not using this particular projection method.

I'm assuming it must be related to my compilation since openMVS work great for others.

Is anyone else observing this behavior?

cdcseacave:
If TPoint2 is returned then the constructor you displayed is not called cause it expects a TPoint3.

paulinus:
Thanks for the explanation. That makes sense. First there is a cast to TPoint2 and then the product between a TPoint2 and `INVERT(q.z)`.

Unfortunately, what i'm seeing on the debugger is that the casting expression
```
((const TPoint2<TYPE>&)q)
```
calls that constructor.

Again, i realize that this must not be happening for anyone else. But i fail to understand why.

cdcseacave:
so try replace `return ((const TPoint2<TYPE>&)q)*INVERT(q.z);` with `return TPoint2<TYPE>(q);`

paulinus:
yes, that is what I did (https://github.com/paulinus/openMVS/commit/f81512073764110e01527ed61f97de80d4b1a149) and it fixes the texturing problem on by side. But i really wonder why it does work for others without that change.

[...]

PR #110, opened 2016-07-29
Do not divide by z twice when projecting.

paulinus:
As commented in #109, on a mac, the conversion from cv::Point3_ to TPoint2 is dividing x and y by z, so there is no need to divide by z a second time.

Creating TPoint2 using numbers instead of a cv::Point3_ avoids dividing twice.


pmoulon:
I tested your PR on gcc Linux, and it works fine for me.

cdcseacave:
Close as it does not apply anymore to latest OpenMVS version. Thank you!

Issue #161, opened 2016-12-18
Floating point exception (core dumped)


charlescva:
### Specifications like the version of the project, operating system, and hardware
Linux, Dell Precision, latest build as of 12/18/2016

### Steps to reproduce the problem
Create NVM from TheiaSfM
InterfaceVisualSFM reconstruction.nvm
DensifyPointCloud reconstruction.mvs
ReconstructMesh reconstruction_dense.mvs

17:21:41 [App     ] Build date: Dec 18 2016, 16:02:04
17:21:41 [App     ] CPU: Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
17:21:41 [App     ] RAM: 31.13GB Physical Memory 19.31GB Virtual Memory
17:21:41 [App     ] OS: Linux 4.7.0-0.bpo.1-amd64 (x86_64)
17:21:41 [App     ] SSE & AVX compatible CPU & OS detected
17:21:41 [App     ] Command line: reconstruction_dense.mvs
17:21:50 [App     ] Scene loaded (8s901ms):
	121 images (121 calibrated) with a total of 26.59 MPixels (0.22 MPixels/image)
	2317330 points, 0 vertices, 0 faces
Floating point exception (core dumped)

Thoughts?

charlescva:
This worked about an hour ago on the same system doing a reconstruct from about 9 photos of a statue.  So I'm not using 130 photos of a building with some background terrain/trees, so the domain may be much larger than my first test.

charlescva:
Okay so just for kicks, I tried opening the Dense PLY reconstruction in MeshLab, and it said that "1 verticies was NaN, Corrected.", and it has loaded in MeshLab successfully. However, trying to export the PLY with color values and re-running it through the ReconstructMesh has failed, probably doesnt like the Meshlab format it pushed out:

root@e817e954b65d:/home/output# ReconstructMesh reconstruction_dense_meshlab.ply 
18:15:11 [App     ] Build date: Dec 18 2016, 16:02:04
18:15:11 [App     ] CPU: Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
18:15:11 [App     ] RAM: 31.13GB Physical Memory 19.31GB Virtual Memory
18:15:11 [App     ] OS: Linux 4.7.0-0.bpo.1-amd64 (x86_64)
18:15:11 [App     ] SSE & AVX compatible CPU & OS detected
[...]

PR #169, opened 2017-01-12
Check Points for NaN

charlescva:
Points which resolve as NaN for any dimension, fail the Delaunay tetrahedralization process in CGAL.  Added check for NaN to at least 1 dimension (X).  More information here, https://github.com/cdcseacave/openMVS/issues/161

### DIFF
--- a/libs/MVS/SceneReconstruct.cpp
+++ b/libs/MVS/SceneReconstruct.cpp
@@ -807,6 +807,7 @@ bool Scene::ReconstructMesh(float distInsert, bool bUseFreeSpaceSupport, unsigne
 		int li, lj;
 		std::for_each(indices.cbegin(), indices.cend(), [&](size_t idx) {
 			const point_t& p = vertices[idx];
+			if (std::isnan(p.hx())) { return; }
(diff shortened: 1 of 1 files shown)


charlescva:
It may also be that we would want to print to the console "point idx# contained NaN... skipping."

cdcseacave:
Are there any NANs? If yes, best would be to find the problem that generates them, instead of adding support for them.

charlescva:
The NaNs came from the dense reconstruction also in the openmvs pipeline, so i agree we should probably address that as well.  However, we should always validate input data for Nulls and such regardless.  In some cases a user may already have a dense cloud and only want to utilize the openmvs reconstruction.  I'll take a look at the densify process when i get a chance.

cdcseacave:
Sounds good, thx Charles!

charlescva:
Thank you for such a great project!

gpanders:
Will this be merged into `master`?

cdcseacave:
Close as it does not apply anymore to latest OpenMVS version. Thank you!

PR #178, opened 2017-02-08
Add snap packaging

mardy:


cdcseacave:
Great contribution! Thank you, Mardy!

mardy:
Hi cdc! Just in case you are wondering how to test the package: make sure you have the `snap` command available; if not, you can follow [these instructions](https://snapcraft.io/docs/core/install).
Once you have `snap` installed, you can just run
```
snap install openmvs
```
And the package will be installed. Note that in order to avoid namespace issues, all command are prefixed with `openmvs`; so, for example, in order to run `ReconstructMesh` you'll actually need to type
```
openmvs.ReconstructMesh
```
On the positive side, one can simply type `openmvs.` and then press TAB twice in the terminal to see the whole list of available OpenMVS tools.

cdcseacave:
The snap packaging recipe is obsolete (targeting Ubuntu 16.04 and defunct dependencies like VCGlib). Dependency management is now standardized on CMake and vcpkg. Closing as obsolete.

PR #291, opened 2018-01-21
minor fix to make the code compile on ubuntu 17.10 again

ThomasEgi:
It does make the compiler happy but it was not extensively tested if it breaks any working code.
A basic reconstruction ran okay, meshing and texture generation was not tested at all.


cdcseacave:
Thank you, fixed by #312

PR #375, opened 2018-10-18
Fix : circumcenter: is not a member of CGAL::Triangulation_cell_base_with_info_3

mlamarre:
Got the error in the title while following Windows build instructions which uses CGAL from vcpkg, it was at version 4.13 when I ran vcpkg install.

The fix is easy but it's not appropriate for previous CGAL versions. Leaving this PR to help. Would require some macros to make work with CGAL 4.2 and up. 

or change build instruction to pick the same version of CGAL in all platforms.


bartoszek:
Add version constraints `#if CGAL_VERSION < 1041301000` so it works with `CGAL<=4.13`.

cdcseacave:
this problem was fixed some time ago, did you try the latest develop code and you still have a problem?

bartoszek:
No, I was using branch master, not develop.

PR #424, opened 2019-04-04
Mesh vertex coloring as an alternative to mesh texturing

mteg:
- Added an option to TextureMesh NOT to generate a texture image but rather assign colors to vertices after seam leveling
- Added vertex color support in Mesh class to be able to hold these colors and save them into a PLY file

This feature is a nice-have when working with large collections of images (say, 500). With the default TextureMesh behavior, the PNG texture tends to take forever to generate and in the end is not loadable anyway. Vertex coloring is an alternative that at least gives the mesh some coloring, even if not that beautiful. Taking the colors from texture patches rather than the dense point cloud benefits from seam leveling, which is very important in case of photos taken under diverse lighting conditions (in my case it is in caves).

### DIFF
--- a/apps/TextureMesh/TextureMesh.cpp
+++ b/apps/TextureMesh/TextureMesh.cpp
@@ -53,6 +53,7 @@ float fOutlierThreshold;
 float fRatioDataSmoothness;
 bool bGlobalSeamLeveling;
 bool bLocalSeamLeveling;
+bool bColorVertices;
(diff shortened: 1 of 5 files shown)


cdcseacave:
Thank you for the contribution! Mesh vertex coloring was integrated into \develop\ in PR #1294 (\Scene::ComputeVertexColors\ and \--vertex-colors\ in \TextureMesh\). Closing as merged.

Issue #447, opened 2019-06-21
ReconstructMesh never end


shauidu:
Hi cdcseacave,

I may have found some bugs in ReconstructMesh project, please check about them.
Problems are in function bool intersect(***);

1. when nb_coplanar == 1, you should check whether there are infinite facets in the output, for the edge may be on the convex boundary. (may have the same problem in case 2, but I did not check that)

2. // skip this cell if the intersection is not in the desired direction
Should we update inter.dist before checking? I think "inter.dist = inter.ray.IntersectsDist(getFacetPlane(in_facet));" should be placed after the following judgement.

cdcseacave:
Please make a PR and add the also the data to reproduce the problem.

shauidu:
I'm not sure whether you can reproduce the problem in the master branch or not, sorry for that. But it do have a problem on my computer.

The data is quite large, it require more than 16GB memory to run, and I'm testing the data on a server using a forked repository, based on v0.8.

I checked the code, and I think it's a logical error, as I said in the first comment. (Maybe I'm wrong :-) )

I will check the data on the master branch later, if I can reproduce the problem, and I'm allowed to share the data. How can I send you the data, it about 1.1GB (200+ images and a densify mvs result).

Thanks

shauidu:
Hi,

I test the same data using master branch code. And it reproduces the problem on one computer, but it works 'fine' on another computer.

The problem is when segment intersects with tetrahedron's edge in bool intersect(***) function in file "openMVS/libs/MVS/SceneReconstruct.cpp", the out_facets will sometimes contain infinite facets when the hit edge is on the boundary.
And inifinite facet's vertex position seems to have different hebavior, sometime it is (0,0,0), sometime one or more its components will be Inf.

I have pulled a request, you can check that.
But I'm not allowed to share the data, sorry for that.

cdcseacave:
Thx, pls check my PR review.

shauidu:
I think the problem is solved, thanks.

PR #450, opened 2019-06-24
intersect_with_infinite_tetrahedron_bug_fix

shauidu:
the detail of this PR is here: "https://github.com/cdcseacave/openMVS/issues/447"

### DIFF
--- a/libs/MVS/SceneReconstruct.cpp
+++ b/libs/MVS/SceneReconstruct.cpp
@@ -606,11 +606,12 @@ bool intersect(const delaunay_t& Tr, const segment_t& seg, const std::vector<fac
 				// the faces in the cells around opposing this common vertex
 				out_facets.clear();
 				struct cell_back_inserter_t {
+					const delaunay_t& Tr;
 					const vertex_handle_t v;
 					const cell_handle_t current_cell;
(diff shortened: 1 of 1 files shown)


cdcseacave:
Thank you for looking into this. I checked the proposed change, but something does not seem right. Here is my concern. At all times, `intersect()` receives only finite faces for intersection check. An finite face means has also finite edges, so how is it possible that you get an intersection with an infinite edge? I trust you have an issue, but maybe this is not it, maybe the problem is somewhere earlier, and not in this function. In that case, it makes sens to fix the root of the problem, and not its symptoms. Unfortunately, I can not help without debug data, as this never happen to me before or anybody else reported a similar issue.

shauidu:
I know what you mean, but I think the current code do not guarantee that "intersect() receives only finite faces".
Here's my opinion. When a segment intersect facet's edge, the out_facets for next loop will be all facets containing this edge. And if the edge is on the boundary of the convex-hull, the out_facets will contain one or more infinite facets.
That's why I add some check for that. Have I missed something? :)

cdcseacave:
it makes sens, I agree
in that case, a similar check is needed for the vertex intersection; I added that too, pls have a look

Issue #438, opened 2019-06-06
Scalable pipeline


cdcseacave:
Add support for processing scenes of any size using a divide & conquer approach:
- [x] break the big SfM reconstruction into smaller chunks (sub-scenes) using some kind of clustering (something like https://github.com/srivathsanmurali/domsetLibrary) with overlapping cameras
- [x] determine exact boundaries between the sub-scenes using SVM (https://docs.opencv.org/3.4/d1/d73/tutorial_introduction_to_svm.html)
- [x] reconstruct each one of them individually using OpenMVS, all the way except texturing
- [x] cut each sub-scene mesh to the boundaries computed by SVM
- [ ] merge them back (at this point the sub-scenes are meshes) using a mesh zipping algorithm (ex. art 5.3 in http://imagine.enpc.fr/publications/papers/PhD011.pdf)
- [ ] apply the texture on each mesh (last module on OpenMVS)

pmoulon:
Seems like a good plan.

You can also look to this paper: `Scalable Surface Reconstruction from Point Clouds with Extreme Scale and Density Diversity, C. Mostegel, R. Prettenthaler, F. Fraundorfer and H. Bischof. CVPR 2017.`

If you want to play with domsetLibrary, we already have it implemented for OpenMVG scenes (see [here](https://github.com/openMVG/openMVG/blob/master/src/software/SfM/clustering/main_ComputeClusters.cpp)).

rjanvier:
@cdcseacave thanks for this roadmap.
I'm a bit unsure about the SVM thing.
for now I assume we use the sparse point clouds from SfM and each point is labeled according to its cluster number, am I right?

cdcseacave:
@pmoulon thx, I didn't read this exact paper yet (but an other one with a similar idea)
[...]

PR #452, opened 2019-06-25
General: Add global ID to MVS::Image

rjanvier:
See #438 (and https://github.com/cdcseacave/openMVS/issues/438#issuecomment-504864399)

- Added the global `ID` to `MVS::Image`
- Added Global ID management in `Scene::LoadInterface(`) and `Scene::SaveInterface() `
- Tried to use the most plausible global `ID` in each interface.

### DIFF
--- a/libs/MVS/Image.h
+++ b/libs/MVS/Image.h
@@ -88,7 +88,7 @@ public:
 	float avgDepth; // average depth of the points seen by this camera
 
 public:
-	inline Image() {
+	inline Image(): ID(NO_ID) {
 		#ifndef _RELEASE
 		width = height = 0;
 		#endif
@@ -124,7 +124,7 @@ public:
 		ar & platformID;
 		ar & cameraID;
 		ar & poseID;
-		if (version > 2) {
+		if (version > 1) {
 			ar & ID;
 		}
 		const String relName(MAKE_PATH_REL(WORKING_FOLDER_FULL, name));
@@ -139,7 +139,7 @@ public:
 		ar & cameraID;
 		ar & poseID;
 		ar & poseID;
-		if (version > 2) {
+		if (version > 1) {
 			ar & ID;
 		}
 		ar & name;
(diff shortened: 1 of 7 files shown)


libs/MVS/Image.h
```diff
@@ -78,6 +78,7 @@ class MVS_API Image
 	uint32_t platformID; // ID of the associated platform
 	uint32_t cameraID; // ID of the associated camera on the associated platform
 	uint32_t poseID; // ID of the pose of the associated platform
+	uint32_t ID; // Global ID of the image
```
cdcseacave: pls initialize it to NO_ID

libs/MVS/Image.h
```diff
@@ -119,21 +120,28 @@ class MVS_API Image
 	#ifdef _USE_BOOST
 	// implement BOOST serialization
 	template<class Archive>
-	void save(Archive& ar, const unsigned int /*version*/) const {
+	void save(Archive& ar, const unsigned int version) const {
 		ar & platformID;
 		ar & cameraID;
 		ar & poseID;
+		if (version > 2) {
```
cdcseacave: this is not enough, you need to inform boost of the current version, ex:
```
BOOST_CLASS_VERSION(MVS::Image, 3)
```
Also, current version is 0, so no need to set it to 3, as 1 is enough.

Pls note that `Interface.h` does not use boost, but a custom serialization in order to be portable. So do not confuse the 2, no connection between them.

rjanvier: thanks a lot! it should be resolved in the new commit.

rjanvier:
Thanks for the review. This issue should be addressed in my last commit.

cdcseacave:
LGTM, feel free to merge

rjanvier:
I think I do not have the right to merge into your branch.

cdcseacave:
hmm, ok, not sure how that works, so I'll just merge it

PR #453, opened 2019-06-28
SceneClustering: Add Initial domset integration

rjanvier:
see #438
this add an initial & basic integration of Domset library into openMVS by creating a SceneClustering binary. This Binary is able to cluster a given OpenMVS Scene and output a set of OpenMVS Scene corresponding to the computed clusters.


libs/MVS/Image.h
```diff
@@ -138,7 +138,6 @@ class MVS_API Image
 		ar & platformID;
 		ar & cameraID;
 		ar & poseID;
-		ar & poseID;
```
rjanvier: this mistake was introduced in the previous commit.

apps/SceneClustering/domset/CHANGES.md
```diff
@@ -0,0 +1,13 @@
+# CHANGES
```
rjanvier: forgot to mention I added overlap

apps/SceneClustering/domset/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+cmake_minimum_required(VERSION 2.8)
```
rjanvier: @cdcseacave: I'm not sure you want to have domset lib inside an `app` directory maybe you want it to be integrated into the `libs` one. Please let me know.

apps/SceneClustering/domset/domset.cc
```diff
@@ -0,0 +1,791 @@
...
+    if (newSP.pos(2) > maxPt.pos(2))
+      maxPt.pos(2) = newSP.pos(2);
+  }
+
+  // finding the number of voxels reqired
+  size_t numVoxelX = static_cast<size_t>(ceil(maxPt.pos(0) - minPt.pos(0)) / sizeX);
+  size_t numVoxelY = static_cast<size_t>(ceil(maxPt.pos(1) - minPt.pos(1)) / sizeY);
+  size_t numVoxelZ = static_cast<size_t>(ceil(maxPt.pos(2) - minPt.pos(2)) / sizeZ);
+
+  /* std::cout << "Max = " << maxPt.pos.transpose() << std::endl;
```
rjanvier: Maybe it could be better if we integrate the bunch of cerr/cout here into your LOG system...

apps/SceneClustering/SceneClustering.cpp
```diff
@@ -0,0 +1,315 @@
...
+				scene_cluster.pointcloud.points.Insert(scene.pointcloud.points[IdxP]);
+				scene_cluster.pointcloud.pointViews.Insert(new_viewArr);
+				if(!scene.pointcloud.colors.IsEmpty())
+					scene_cluster.pointcloud.colors.Insert(scene.pointcloud.colors[IdxP]);
+				if(!scene.pointcloud.pointWeights.IsEmpty())
+					scene_cluster.pointcloud.pointWeights.Insert(scene.pointcloud.pointWeights[IdxP]);
+			}
+		}
+		LOG(String::FormatString("Saving cluster #%i", i));
+		scene_cluster.Save(baseFileName + String::FormatString("_cluster_%i.mvs", i), (ARCHIVE_TYPE)OPT::nArchiveType);
```
rjanvier: here I choose convention over configuration. I do not give user the choice of the directory/suffix.

apps/SceneClustering/domset/CMakeLists.txt
cdcseacave: I'd usually add libs in `libs` folder, for example domset could go in `Math`; however in this case this lib is very probable to be used only for this specific task

in conclusion feel free to add it where you think is better; just note that is you add it in the app, you can not use it inside MVS lib, so if you think u can do all u need to do outside MVS in app is just fine

apps/SceneClustering/domset/domset.cc
cdcseacave: indeed, for that you just need to include Common/Common.h and use GET_LOG()

apps/SceneClustering/SceneClustering.cpp
cdcseacave: it is fine

apps/SceneClustering/SceneClustering.cpp
```diff
@@ -0,0 +1,315 @@
...
+	uint32_t curr_ID = 0;
+	FOREACH(IdxC, scene.images)
+	{
+		const auto & curr_image = scene.images[IdxC];
+		if(curr_image.IsValid())
+		{
+			view_fwd_reindexing[IdxC] = curr_ID;
+			view_bkwd_reindexing[curr_ID] = IdxC; 
+			nomoko::View v;
+			v.rot = Eigen::Matrix<double,3,3,1>(curr_image.camera.R).cast<float>();
```
cdcseacave: I'd use REAL instead of double

apps/SceneClustering/SceneClustering.cpp
```diff
@@ -0,0 +1,315 @@
...
+		;
+
+	// group of options allowed both on command line and in config file
+	boost::program_options::options_description config("Texture options");
+	config.add_options()
+		("input-file,i", boost::program_options::value<std::string>(&OPT::strInputFileName), "Input filename containing camera poses and image list")
+		("voxel-size,x", boost::program_options::value<float>(&OPT::fVoxelSize)->default_value(5.f), "Size of a cell in the voxel grid: level of simplification of the original point cloud")
+		("min-cluster-size,m", boost::program_options::value<unsigned>(&OPT::nMinClusterSize)->default_value(30), "Min number of camera in a cluster" )
+		("max-cluster-size,M", boost::program_options::value<unsigned>(&OPT::nMaxClusterSize)->default_value(50), "Max number of camera in a cluster")
+		("cluster-overlap,o", boost::program_options::value<unsigned>(&OPT::nClusterOverlap)->default_value(4), "Number of views in overlap [NOT implemented yet]" )
```
cdcseacave: I'd rather use a percentage here

apps/SceneClustering/domset/nanoflann.hpp
```diff
@@ -0,0 +1,2039 @@
+/***********************************************************************
```
cdcseacave: however I would definitely place this file in Math lib, as this can easily be used in many other places

cdcseacave:
windows build fails cause it has a low OpenMP support, pls use `int` for any omp for loops

apps/SceneClustering/domset/CHANGES.md
rjanvier: fixed in 382bc4a16a9a149f37ad1d5c7e8718fb881ec759

apps/SceneClustering/domset/domset.cc
rjanvier: thanks! fixed in 6e431611556137bcec6e3ecdb2f12c2a29b24fd8

apps/SceneClustering/SceneClustering.cpp
rjanvier: thanks! fixed in 9e84c2fcc0f16d451483c30359da754c5eab2286

apps/SceneClustering/SceneClustering.cpp
rjanvier: handled in 382bc4a16a9a149f37ad1d5c7e8718fb881ec759

apps/SceneClustering/domset/nanoflann.hpp
rjanvier: resolved by bccdbe40c9e9a7eeb84811bb484d56eed82fd51d

rjanvier:
Hey Dan thanks a lot for your detailed review!
I think I have addressed all your concern in the bunch of commit I made this morning.
Nevertheless it seems there is another VS error and I can't figure out what's going on. Would you mind have a quick look please?

cdcseacave:
indeed, that is a bit tricky, when u include nanoflann.h, pls use this:
```
#pragma push_macro("malloc")
#undef malloc
#pragma push_macro("PI")
#undef PI
#include "nanoflann.h"
#pragma pop_macro("PI")
#pragma pop_macro("malloc")
```

rjanvier:
thx a lot, done in last commit.

PR #454, opened 2019-06-29
Add depth maps fusion at the cluster level

rjanvier:
see #438 
In order to take advantage of the Affinity propagation clustering. This PR would allows to :

- [x] Disable the fusion step in the DensifyPointCloud binary in order to compute and filter depth maps within the main Scene and avoid out-of-memory errors
- [x] Correct the localID/globalID mechanism as suggested here https://github.com/cdcseacave/openMVS/pull/454#issuecomment-507972989
- [x] Add ability to run the fusion on each clusters Scene independently in DensifyPointCloud

### DIFF
--- a/libs/MVS/DepthMap.cpp
+++ b/libs/MVS/DepthMap.cpp
@@ -72,7 +72,6 @@ MDEFVAR_OPTDENSE_uint32(nMinViewsFilterAdjust, "Min Views Filter Adjust", "minim
 MDEFVAR_OPTDENSE_uint32(nMinViewsTrustPoint, "Min Views Trust Point", "min-number of views so that the point is considered for approximating the depth-maps (<2 - random initialization)", "2")
 MDEFVAR_OPTDENSE_uint32(nNumViews, "Num Views", "Number of views used for depth-map estimation (0 - all views available)", "0", "1", "4")
 MDEFVAR_OPTDENSE_bool(bFilterAdjust, "Filter Adjust", "adjust depth estimates during filtering", "1")
-MDEFVAR_OPTDENSE_bool(bFuse, "Fuse", "fuse depth maps into a point cloud", "1")
 MDEFVAR_OPTDENSE_bool(bAddCorners, "Add Corners", "add support points at image corners with nearest neighbor disparities", "1")
 MDEFVAR_OPTDENSE_float(fViewMinScore, "View Min Score", "Min score to consider a neighbor images (0 - disabled)", "2.0")
 MDEFVAR_OPTDENSE_float(fViewMinScoreRatio, "View Min Score Ratio", "Min score ratio to consider a neighbor images", "0.3")
--- a/libs/MVS/DepthMap.cpp
+++ b/libs/MVS/DepthMap.cpp
@@ -72,6 +72,7 @@ MDEFVAR_OPTDENSE_uint32(nMinViewsFilterAdjust, "Min Views Filter Adjust", "minim
 MDEFVAR_OPTDENSE_uint32(nMinViewsTrustPoint, "Min Views Trust Point", "min-number of views so that the point is considered for approximating the depth-maps (<2 - random initialization)", "2")
 MDEFVAR_OPTDENSE_uint32(nNumViews, "Num Views", "Number of views used for depth-map estimation (0 - all views available)", "0", "1", "4")
 MDEFVAR_OPTDENSE_bool(bFilterAdjust, "Filter Adjust", "adjust depth estimates during filtering", "1")
+MDEFVAR_OPTDENSE_bool(bFuse, "Fuse", "fuse depth maps into a point cloud", "1")
 MDEFVAR_OPTDENSE_bool(bAddCorners, "Add Corners", "add support points at image corners with nearest neighbor disparities", "1")
 MDEFVAR_OPTDENSE_float(fViewMinScore, "View Min Score", "Min score to consider a neighbor images (0 - disabled)", "2.0")
 MDEFVAR_OPTDENSE_float(fViewMinScoreRatio, "View Min Score Ratio", "Min score ratio to consider a neighbor images", "0.3")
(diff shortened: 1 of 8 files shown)


libs/MVS/SceneDensify.cpp
```diff
@@ -1737,37 +1737,42 @@ bool Scene::DenseReconstruction()
...
-			case 1:
-				++nPoints1m;
-				break;
-			case 2:
-				++nPoints2;
-				break;
-			default:
-				++nPoints3p;
+	// fuse all depth-maps if needed
+	if(OPTDENSE::bFuse)
```
cdcseacave: I think it will look more clean if let the code as it was, and just add `return true` after if

libs/MVS/DepthMap.cpp
```diff
@@ -72,6 +72,7 @@ MDEFVAR_OPTDENSE_uint32(nMinViewsFilterAdjust, "Min Views Filter Adjust", "minim
 MDEFVAR_OPTDENSE_uint32(nMinViewsTrustPoint, "Min Views Trust Point", "min-number of views so that the point is considered for approximating the depth-maps (<2 - random initialization)", "2")
 MDEFVAR_OPTDENSE_uint32(nNumViews, "Num Views", "Number of views used for depth-map estimation (0 - all views available)", "0", "1", "4")
 MDEFVAR_OPTDENSE_bool(bFilterAdjust, "Filter Adjust", "adjust depth estimates during filtering", "1")
+MDEFVAR_OPTDENSE_bool(bFuse, "Fuse", "fuse depth maps into a point cloud", "1")
```
cdcseacave: if you add a command line param for fusion, no need to add it also as a INI param, simply pass it as a variable to densification fucntion

libs/MVS/SceneDensify.cpp
rjanvier: Thanks, agree!

libs/MVS/DepthMap.cpp
rjanvier: Ok thank you. I was not sure. I was just trying to be exhaustive here.

rjanvier:
--keep-depth-maps argument does not exists in the current code base. My initial though was to add something above this code https://github.com/cdcseacave/openMVS/blob/c70f51f2c87b4cb6548084067571746d16358d2a/libs/MVS/SceneDensify.cpp#L1808-L1813
like that for example, given a putative `--fuse-cluster-scene` arg:
```cpp			
if(OPTDENSE::bFuseClusterScene && File::access(ComposeDepthFilePath(images[idx].ID, "dmap"))) {
  if (!depthData.Load(ComposeDepthFilePath(images[idx].ID, "dmap"))) {
    VERBOSE("error: invalid depth-map '%s'", ComposeDepthFilePath(images[idx].ID, "dmap").c_str());
    exit(EXIT_FAILURE);
  }
break;
}
```

cdcseacave:
I made a mistake, no need for any param, simply run densification for each cluster when the `dmap` files exist and all will work as expected (depth-maps are loaded during fusion)
just make sure `OPTDENSE::nOptimize` is set to 0 for all cluster densification runs, as we do not want to duplicate those postprocessings

rjanvier:
I’m not sure (but I can’t check right now), we have to refer to the global ID to load the good depth map for each image. 
For example, let say first image in cluster A is image number 455 in the non clustered scene. If we want to fuse cluster A without code modification it will load depth0001.dmap instead of depth0455.dmap
So I think we need a flag to tell DensifyPointCloud to use image::ID in order to compute depth map name instead of positional indexes.

cdcseacave:
u r totally right, but we do not need any special flag, just modify the code to always use the global ID instead of the local one

rjanvier:
Ok. My concern on this topic is that one can use a base Scene file without global ID ( here "base" = "non clustered", the "main Scene") either because they can use an old Scene file (created before the introduction of the global ID) or either an old importer (for example for now the openMVS2openMVG importer still do not use global ID, I plan tu update it but I'm not sure it will make it for the next release). so two solution here
- Use global ID everywhere like you suggest but in this case I thing we should provide a binary to fill global IDs into a Scene file when its needed (ol importer/old scene). It's the simpler way but we'll break backward compatibility.
- The current paradigm I choose (To fully understand it you'll have to forget I have updated all the interfaces to export global IDs, in this case it's a mistake) In this paradigm global ID is an internal mechanism to openMVS and when it's used, it always make reference to the index of an image in the imageArr of the base Scene file. this way backward compatibility is guarantied but we need aditional tricks (like this specific flag during the fusion) to make sure to use globals ID instead of imageArray indexes when needed.
Let me know what do you think. Thanks!

cdcseacave:
How about each time a MVS project is loaded to make sure the globalID is valid; and if not just set the localID?

rjanvier:
yeah, it seems to be the best plan!

libs/MVS/DepthMap.cpp
rjanvier: Ok, I'm finally not sure of what to fix here. Densification function is not taking any argument (on the contrary to Refine and Mesh functions for example). Should I remove this line?

cdcseacave: I was trying to suggest to add it as a param to `bool Scene::DenseReconstruction()` function, but u do as u think is better

rjanvier: Well, I want to maximize the chances to see this code merged into master one day ;). I'm not totally against the parameter, I just wanted to be consistent with existing code.

libs/MVS/SceneDensify.cpp
rjanvier: @fixed via 186e8f1b40ef118eb54f528da1c63a66f394a09a

libs/MVS/DepthMap.cpp
rjanvier: I hope I understood well and I have implemented the method parameter in this commit 813bde73b5b3ff04a149671c811a498e64611ebd so feel free to pick the one you prefer.

PR #460, opened 2019-07-09
(DensifyPointCloud) Fix crash when working directory is provided 

### LINKED ISSUE #445: DensifyPointCloud crashes when -w (working folder) argument is used
### Specifications like the version of the project, operating system, and hardware

* Current master
* Ubuntu 16.04 with updates
* Xeon E3-1246v3, 32GB

### Steps to reproduce the problem

When running DensifyPointCloud use -w argument ie. ```DensifyPointCloud scene.mvs -w /datasets/mvs```

openmvs

The breaking change was introduced in [this commit](https://github.com/cdcseacave/openMVS/commit/e017ff346ff2c809466445ee229ee00ed34a257e).

Test material:
[scene.mvs.zip](https://github.com/cdcseacave/openMVS/files/3295489/scene.mvs.zip)
[imgs.zip](https://github.com/cdcseacave/openMVS/files/3295490/imgs.zip)

csparker247:
Fixes an exception that was being thrown when the working directory option `-w` is provided. 

Problem: `MAKE_PATH` appends working directory to dense config file path (`OPT::strDenseConfigFileName`). When this path is empty (which is the default) and when working directory is provided, the config loader effectively tries to load the working directory path as a config file. This fails deep inside the config loader.

Solution: Make sure the config file path is non-empty before appending the working directory and trying to load the config info.

Fixes #445

### DIFF
--- a/apps/DensifyPointCloud/DensifyPointCloud.cpp
+++ b/apps/DensifyPointCloud/DensifyPointCloud.cpp
@@ -167,7 +167,7 @@ bool Initialize(size_t argc, LPCTSTR* argv)
 		OPT::strOutputFileName = Util::getFileFullName(OPT::strInputFileName) + _T("_dense.mvs");
 
 	// init dense options
-	if (!Util::isFullPath(OPT::strDenseConfigFileName))
+	if (!OPT::strDenseConfigFileName.IsEmpty() && !Util::isFullPath(OPT::strDenseConfigFileName))
 		OPT::strDenseConfigFileName = MAKE_PATH(OPT::strDenseConfigFileName);
 	OPTDENSE::init();
 	const bool bValidConfig(OPTDENSE::oConfig.Load(OPT::strDenseConfigFileName));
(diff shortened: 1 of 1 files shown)


cdcseacave:
I see how this change can make things more clear, but it is not needed.
Also, even if `-w` is not defined, the working directory is still define by current folder, so using or not `-w` makes zero difference in that section.

csparker247:
> I see how this change can make things more clear, but it is not needed.

I get an exception without this change (see #445). I've confirmed that the changes you made also avoid the exception.

> Also, even if `-w` is not defined, the working directory is still define by current folder, so using or not `-w` makes zero difference in that section.

Working directory is current directory by default because `WORKING_FOLDER` (which is what gets appended to all paths) is empty by default. Running `DensifyPointCloud -i scene.mvs` and `DensifyPointCloud -i scene.mvs -w /full/path/to/current/directory/` should be the same, but the latter will cause an exception without this PR.

PR #462, opened 2019-07-26
Adding re-texturing option:

thunders82:
allows to backup and reuse the atlas to re-texture the mesh later using 
the same UV's


thunders82:
Sorry for the noise I created. I'm learning Atom+git+github the hard way I suppose.
I'm closing this pull request until I understood how to fix this windows compilation issue without having any windows computers around :) Cheers

thunders82:
re-opening for testing

thunders82:
Alright. It looks like everything is compiling properly on win32 now. Sorry again for the noise...

thunders82:
The 2 last commits gives the ability to read 32bit EXR files and load the data as float 32bit (linear) or 8bit (log) output.

thunders82:
Last commit is quick conversion to float/32F. TextureMesh has been tested and seems to work as expected.

cdcseacave:
Sry for the late reply!

This changes look very professional, it does not look like you are only starting to learn programming.

Here are some suggestions:

I like a lot the idea of using Image32F3 instead of Image8U3 in order to support all types of inputs. It is very simple and should just work. The only downside is that this will increase enven farther the memory requirements. It would be best if we can use a dynamic image type, that stores pixels in uint8_t or float depending on the input image. Or at least use a typedef that can easily be changed at compile time (so instead of manually replacing all instances of Image8U3 to Image32F3, to make a `typedef Image32F3 InputImage` that can easy be changed back to Image8U3 and use InputImage everywhere.

To make your life easier on Windows, use VCPKG to install and compile 3rd party libs. Not only that will save you time to manage all those libs, but should make it easy also to find them using CMake in a consistent way (no need for custom FindPackage).

This PR grew very large, and targets two different problems. Pls split it in 2 PR, one for each problem.

Thank you for the HARD work.

cdcseacave:
Thank you for the effort on this PR. Overhauling the entire texture pipeline to 32-bit float quadruples memory consumption and touches 31 files. Closing this PR as abandoned.

Issue #303, opened 2018-02-21
ReconstructMesh crash


kahlan78:
Hi,
running the ReconstructMesh step (on Win10), it crashes with the following error:

ReconstructMesh.exe model.mvs -f 1 
Points inserted 100791 (93.48%, 800ms, ETA 55ms)...GNU MP: Cannot allocate memory (size=52776842297400)

If i add this option --min-point-distance with high values (i.e. 20) the step doesn't crash anymore. With another dataset, i get the opposite: with the default value of --min-point-distance the step works fine, instead with --min-point-distance 20 the step crashed with GNU MP: Cannot allocate memory.

Is there a link between this option and the crash? How do i choose the correct value of the--min-point-distance? What the meaning of this option?

Tks

cdcseacave:
I can not debug the issue with only this. Pls provide the crashing dataset, project, command line, etc

givanfp:
I don't know if this is the same issue but RecontstructMesh crashes too. 
Code from master, compiled on Windows 10 /64-bits staticlibs, ReconstructMesh running without parameters, only the input scene.

Error:
<code>
Exception thrown: read access violation.
m was 0xFFFFFFFFFFFFFFFF.
</code>

``
The error is on the line 393 in common\Plane.inl

```
void TFrustum<TYPE,DIMS>::Set(const MATRIX3x4& m, TYPE w, TYPE h, TYPE n, TYPE f)
{
	MATRIX4x4 M(MATRIX4x4::Identity());
	M.template topLeftCorner<3,4>() = m;  <<==here throws the exception
	Set(M, w, h, n, f);
} // Set
```


Callstack:
```
ReconstructMesh.exe!SEACAVE::TFrustum<double,4>::Set(const Eigen::Matrix<double,3,4,1,3,4> & m, double w, double h, double n, double f) Line 393	C++
[...]

PR #472, opened 2019-09-11
Fix ReconstructMesh crash #303

flgw:
Eigen>=3.3.x segfaults on Linux with gcc

Using dynamic (runtime) version of topLeftCorner works


cdcseacave:
Thx for this, but I think this is a workaround rather than a real fix, just waiting to bread with the next eigen version. I real fix should be making sure that all vectors and matrices used are aligned as per Eigen requirements. This is I think the real reason for crashing on linux.

flgw:
Yeah, sure. Just thought, since this bug is already been around for more than a year, that a working workaround might be valuable. At least until a proper solution is found.

May I ask what you concerns are in particular? Is it performance you are worried about?

cdcseacave:
addressed in develop

PR #476, opened 2019-09-18
Expose SceneDensify data structures and functions

indianajohn:
This is a very small refactoring of the MVS library to:

* Allow the computation of depth maps without the fusion step, in a library call.
* Allow optional inspection of the `DenseDepthMap` data structure following densification

Let me know if this doesn't fit with your vision for the library and we can maintain these changes on our fork only.


libs/Common/Types.inl
```diff
@@ -2147,7 +2147,11 @@ void TImage<TYPE>::toGray(TImage<T>& out, int code, bool bNormalize, bool bSRGB)
 		out.create(rows, cols);
 	ASSERT(cv::Mat::isContinuous());
 	ASSERT(out.cv::Mat::isContinuous());
+#ifdef __clang__
 	const int scn(cv::Mat::channels());
+#else
```
indianajohn: GCC 7.4.3 was throwing an error on this line.

```
libs/MVS/../Common/Types.inl:2150:33: error: cannot call member function 'int cv::Mat::channels() const' without object
```
I suppose this could be a GCC version problem, OpenCV version problem, or something else entirely. I'm just putting this in here now as something that solves the problem in our case, and solicit your advice. We can refine this `#ifdef` to be something closer to the actual trigger or remove this from the PR as out of scope if you would like.

libs/MVS/SceneDensify.h
```diff
@@ -0,0 +1,103 @@
...
+	Image8U::Size prevDepthMapSize; // remember the size of the last estimated depth-map
+	Image8U::Size prevDepthMapSizeTrg; // ... same for target image
+  DepthEstimator::MapRefArr coords; // map pixel index to zigzag matrix coordinates
+  DepthEstimator::MapRefArr coordsTrg; // ... same for target image
+};
+
+struct MVS_API DenseDepthMapData {
+  Scene& scene;
+  IIndexArr images;
+  IIndexArr neighborsMap;
```
pmoulon: You seem to have a tab alignment issue. It could be fixed by your text/code editor.
You have some tabs with 2 or 4 spaces.

libs/MVS/SceneDensify.h
```diff
@@ -0,0 +1,103 @@
...
+	DepthMapsData detphMaps;
+	volatile Thread::safe_t idxImage;
+	SEACAVE::EventQueue events; // internal events queue (processed by the working threads)
+	Semaphore sem;
+	CAutoPtr<Util::Progress> progress;
+
+	DenseDepthMapData(Scene& _scene)
+		: scene(_scene), detphMaps(_scene), idxImage(0), sem(1) {}
```
pmoulon: typo `detphMaps` -> `depthMaps`

indianajohn: It's spelled `detphMaps` in the original code, and also in various places in `SceneDensify.cpp`. I can change it everywhere if folks want - I was just looking to keep this change set scoped to what I was trying to do, so I left it in on purpose.

libs/MVS/SceneDensify.h
indianajohn: My editor uses spaces, the repo seems to have standardized on tabs.

I'll fix it up.

indianajohn: I replaced instances of leading spaces with leading tabs in all cases, as seems to be the style of the repository. Resolving this comment.

libs/Common/Types.inl
cdcseacave: try using `this->` like `this->cv::Mat::channels()`

libs/MVS/SceneDensify.h
cdcseacave: it is a typo, pls feel free to fix it everywhere, thx

libs/MVS/SceneDensify.cpp
```diff
@@ -29,8 +29,7 @@
 *      containing it.
 */
 
-#include "Common.h"
-#include "Scene.h"
+#include "SceneDensify.h"
```
cdcseacave: this might not work on Windows
"Common.h" is a precompiled header and needs to be always added

libs/MVS/SceneDensify.h
```diff
@@ -0,0 +1,103 @@
...
+*      that material or in the Appropriate Legal Notices displayed by works
+*      containing it.
+*/
+
+#ifndef _MVS_SCENEDENSIFY_H_
+#define _MVS_SCENEDENSIFY_H_
+
+// I N C L U D E S /////////////////////////////////////////////////
+
+#include "Common.h"
```
cdcseacave: and should not be added to header files

libs/MVS/SceneDensify.h
```diff
@@ -0,0 +1,103 @@
+/*
+* DepthMap.h
```
cdcseacave: name

libs/MVS/SceneDensify.cpp
indianajohn: Okay, I'll check and test compilation on Windows. Thanks.

libs/MVS/SceneDensify.h
indianajohn: Fixed.

libs/MVS/SceneDensify.h
indianajohn: Changed this.

libs/Common/Types.inl
indianajohn: I changed to this and it worked on all of our platforms, thanks.

libs/MVS/SceneDensify.cpp
indianajohn: After changing the include structure I was able to build OpenMVS on Windows.

libs/MVS/SceneDensify.h
indianajohn: Addressed above ^^^

cdcseacave:
@indianajohn I added small changes regarding formatting, and proposed a slight change in the densification function call: I removed the empty function; that should not impact the functionality; in the future if you need access to fusion module as well, we will have to move the remaining half in a new function

pls tell me if this is fine with you

indianajohn:
I was able to make these changes work well on my end. Thanks.

cdcseacave:
So all good with you to merge, or do you plan new changes?

indianajohn:
What's on this branch is useful to us as-is - let's merge it if you're okay with it.

cdcseacave:
pls point it to develop instead of master

apps/DensifyPointCloud/DensifyPointCloud.cpp
```diff
@@ -180,7 +180,7 @@ bool Initialize(size_t argc, LPCTSTR* argv)
 	OPTDENSE::nOptimize = nOptimize;
 	OPTDENSE::nEstimateColors = nEstimateColors;
 	OPTDENSE::nEstimateNormals = nEstimateNormals;
-	if (!bValidConfig && !OPT::strDenseConfigFileName.IsEmpty())
+	if (!bValidConfig)
```
cdcseacave: why is that needed? it was added in a recent issue causing crashes in certain scenarios

indianajohn: It isn't really - I think your window might have been stale when you looked at this or something. That was in this PR after I merged in develop, but I noticed it and reverted the irrelevant changes.

cdcseacave: strange, all changes were made on github

PR #490, opened 2019-10-21
Fix API of pca_estimate_normals for CGAL version > 4.13

maxGimeno:
Hello,
We are about to upgrade CGAL version on vcpkg to 5.0, and in that version some deprecated code has been removed, which causes the build of openMVS to fail. I propose this PR to fix the calls for recent versions of CGAL, and in parallel I propose [a patch](https://github.com/microsoft/vcpkg/pull/8659/files#diff-9046912f3fa504b1f13c5e717992fd8b) for vcpg in [my PR](https://github.com/microsoft/vcpkg/pull/8659).


cdcseacave:
thank you!

Issue #404, opened 2019-01-21
Segfault with ReconstructMesh


fishcu:
### Specifications like the version of the project, operating system, and hardware
Ubuntu 16.04, CUDA 10.0, OpenCV 4.0.0-rc built from source.
32GB of RAM, 12GB of VRAM.

OpenMVS is latest from master branch.

### Steps to reproduce the problem
I'm trying to reconstruct a mesh. Same dataset as in #402. Dense reconstruction finished successfully.
Here's the full output:
```
bin/ReconstructMesh -w /home/isaac/Downloads/buddha_recon_2 '/home/isaac/Downloads/buddha_recon_2/scene_dense.mvs' 
Eigen version 3.2.10 is used!
15:38:48 [App     ] Build date: Jan 21 2019, 15:07:48
15:38:48 [App     ] CPU: Intel(R) Core(TM) i7-6800K CPU @ 3.40GHz
15:38:48 [App     ] RAM: 31.33GB Physical Memory 977.00MB Virtual Memory
15:38:48 [App     ] OS: Linux 4.15.0-43-generic (x86_64)
15:38:48 [App     ] SSE & AVX compatible CPU & OS detected
15:38:48 [App     ] Command line: -w /home/isaac/Downloads/buddha_recon_2 /home/isaac/Downloads/buddha_recon_2/scene_dense.mvs
15:38:48 [App     ] MapSMtoCores for SM 6.1 is undefined; default to use 128 cores/SM
15:38:48 [App     ] CUDA device 0 initialized: TITAN Xp (compute capability 6.1; memory 11.91GB)
15:38:58 [App     ] Scene loaded (10s398ms):
	41 images (41 calibrated) with a total of 235.27 MPixels (5.74 MPixels/image)
	11343837 points, 0 vertices, 0 faces
Points inserted 11343837 (100%, 1m5s432ms)           
Segmentation fault (core dumped)
```
(I added the eigen version line for debug purposes.)

I saw on #303 that you might need to use an exact Eigen version, so I use the one specified in build instructions.

**Any ideas how to fix this? I should have plenty of RAM available.**

For reference, here's an `ldd` output to show other libraries linked.
```
ldd bin/ReconstructMesh 
	linux-vdso.so.1 =>  (0x00007ffd95119000)
	libboost_iostreams.so.1.58.0 => /usr/lib/x86_64-linux-gnu/libboost_iostreams.so.1.58.0 (0x00007fb566ba6000)
[...]

PR #494, opened 2019-10-22
Fix segmentation fault on Scene Refine Cuda

BhanukaUOM:
Fix For Issue #404
Fixed "segmentation fault (core dumped)" on CUDA Reconstruct Mesh.


cdcseacave:
sry, this is not a fix, but masking out a problem
`idxFace` can not be negative as it is unsigned, and can not be bigger then the number of faces; if it is, that is the problem to be fixed, and not this hiding the problem fix

Issue #426, opened 2019-04-17
BoundaryWeight is not found in VGC


skinkie:

cdcseacave:
no info, at least the decency to fill in the template

skinkie:
The version current git. The problem is in:
@@ -997,7 +997,7 @@ void Mesh::Clean(float fDecimate, float fSpurious, bool bRemoveSpikes, unsigned

PR #503, opened 2019-11-25
Parameter name change to reflect current state of vcglib's master

prass-anyvision:
Build breaks because without this change. Vcglib does not contain the previous parameter anymore.


prass-anyvision:
Actually mentioned in #426

cdcseacave:
thank you, but I can not merge this because it does not maintain compatibility with the old version of VCG

prass-anyvision:
Your builds for newer versions are broken though. Or, you have to mention a specific version of VCG in the build wiki

cdcseacave:
VCGlib has been completely removed from OpenMVS in favor of CGAL and halfmesh. Closing as obsolete.

PR #504, opened 2019-12-05
Add docker functionality

tjdahlke:
I had trouble installing openMVS on my local machine, so I used Docker to make a container that would build/install it and simplify the use of the the binaries by others. A Dockerfile has been added so that users can build the container from scratch, and a runMe.sh script and README.md have also been added for ease of use.

This makes the build environment standardized and reproducible, and allows the work in this repo to be accessible to a wider range of users.


docker/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+RUN apt update
+RUN apt install -y cmake git vim
+
+# Prepare and empty machine for building:
+RUN apt-get update -yq
+RUN apt-get install -yq
+RUN apt-get -y install git mercurial cmake libpng-dev libjpeg-dev libtiff-dev libglu1-mesa-dev
+
+# Eigen
+RUN hg clone https://bitbucket.org/eigen/eigen#3.2
```
pmoulon: From the repo title: ```This is a deprecated repository kept as is for convenience.
 Eigen's official repository have been moved to https://gitlab.com/libeigen/eigen```

docker/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+RUN git clone https://github.com/cdcseacave/VCG.git vcglib
+
+# Build from stable release openMVS1.0
+RUN git clone https://github.com/cdcseacave/openMVS.git --branch v1.0
+
+# Uncomment below (and comment above) to use the latest commit from the master branch
+# RUN git clone https://github.com/cdcseacave/openMVS.git openMVS
+
+RUN mkdir openMVS_build
+RUN echo "hello"
```
pmoulon: Do we want to keep this log?

tjdahlke: Whoops! Fixed.

docker/Dockerfile
tjdahlke: Good catch; looks like they migrated it to gitlab just yesterday. Corrected and tested.

cdcseacave:
LGTM but did you actually run the code in docker and reconstructed some datasets?
I ask because I see you use Eigen 3.3.7, which from what I know does not work ok on UBUNTU with OpenMVS

tjdahlke:
@cdcseacave, I rebuilt the Docker container from scratch (from the most recent Dockerfile using ubuntu:18.04 and Eigen3.3.7) and tested running DensifyPointCloud using some openMVG SFM results as input. I was able to successfully build a dense point cloud; and it looks great. I have not yet tested the other programs. Is there a specific test I should run to check for the issue you're referring to?

cdcseacave:
no, I do not use linux, but users were complaining about crashes, if you run all modules fine it is ok

pmoulon:
I use multiple time OpenMVG in this docker build https://github.com/rennu/dpg with success.
It's gonna be nice to have an official docker image attached to the repo.

cdcseacave:
@pmoulon you probably meant OpenMVS, but that docker image uses Eigen 3.2 as I recommend

tjdahlke:
To clarify @pmoulon, the Docker image will not be attached to the repo (too big of a file). However, it might make sense to make an openMVS dockerHub account (free) to host the official docker image. That way, a user doesn't have to make the docker image from scratch (although they still could if they wanted to using the Dockerfile), but instead they could just run something like:

`docker pull openMVS/openmvs-ubuntu:latest`

and just pull an 'official' image. I have such an image hosted publicly at [tjdahlke/openmvs-ubuntu](https://hub.docker.com/repository/docker/tjdahlke/openmvs-ubuntu) right now, but might be better to have an official openMVS dockerHub account to host that instead. I would be happy to set it up / add you both as collaborators (or let you set it up and upload the image as a collaborator).

cdcseacave:
It would be of great help if you can set it up, you clearly have the expertise. Pls do not ignore though the Eigen version suggestion: either test that all works, or use v3.2

pmoulon:
@tjdahlke Sorry I meant have the Dockerfile in the main OpenMVS repo would be nice to let anyone build and manage their custom build if they need.
Also as you said, be able to pull a prebuilt image is a nice feature.

tjdahlke:
@cdcseacave I have not ignored the eigen version suggestion, I just didn't get a chance to do further testing till today. I find that while **DensifyPointCloud** works fine with eigen v3.3.7, **ReconstructMesh** does not for some reason (seg fault). Because of this, I am reverting back to eigen v3.2.1, which is the first listed 3.2.~ series release. I tested all programs using the eigen v3.2.1 release and they all worked fine.

tjdahlke:
@cdcseacave and @pmoulon; No problem. I will set up a DockerHub account for openMVS soon then.

tjdahlke:
openMVS DockerHub account set up / integrated into updated scripts. Should be ready for merge; please let me know if there is anything else I can add.

cdcseacave:
thx @tjdahlke , but pls use 3.2 branch with all latest bug fixes, it is known of working fine any 3.2.x Eigen

tjdahlke:
@cdcseacave Updated to use the 3.2 branch.

tjdahlke:
@cdcseacave @pmoulon Anything else missing here? Otherwise should be ready to merge.

cdcseacave:
I did not test this, and I'll not have a chance to, so I'll commit based on your tests. Thank you for your contribution!

cdcseacave:
@tjdahlke as you do have experience with linux builds, do you have any idea why the AppVeyor builds fail in linux for the automatic build? I do not see any error reported, and on local linux machine all works fine.

PR #517, opened 2020-01-15
fix zero index normals in obj export

tompollok:
That should output obj files without 0 indices for normals in case there exists no normals.
It seems like i dont have to modify the obj reader as you check for the case of only vertices and texture coordinate indices.


libs/IO/OBJ.cpp
```diff
@@ -164,8 +166,10 @@ bool ObjModel::Save(const String& fileName, unsigned precision, bool texLossless
 			out << "f";
 			for (size_t k = 0; k < 3; ++k) {
 				out << " " << face.vertices[k]  + OBJ_INDEX_OFFSET
-					<< "/" << face.texcoords[k]  + OBJ_INDEX_OFFSET
-					<< "/" << face.normals[k]  + OBJ_INDEX_OFFSET;
+					<< "/" << face.texcoords[k]  + OBJ_INDEX_OFFSET;
+				if(hasNormals) {
```
cdcseacave: I think it is more general if using `if (face.normals[k]  != NO_ID)`, same for `texcoords`

tompollok: The problem is that i dont know if the specification allows for mixing faces.

These are ok:
f v1 v2 v3 ....
f v1/vt1 v2/vt2 v3/vt3 ...
f v1/vt1/vn1 v2/vt2/vn2 v3/vt3/vn3 ...
f v1//vn1 v2//vn2 v3//vn3 ...

But i dont know if mixes like the following are ok which could happen if its done individually per face:

f v0 v1//vn1 v2/vt2/vn2 v3/vt3 ...

Thats why i just solved the case when there are no normals at all.

cdcseacave:
thx, I addressed this in 3ae6aa4 slightly different, pls use develop

PR #537, opened 2020-03-06
Added support to import binary colmap files

SBCV:
This PR allows to directly import binary colmap files (i.e. dense colmap reconstructions).
This has several advantages:
- Easier to use (no need to create TXT files by hand) 
- BIN files store values with higher accuracy
- Import of BIN files is slightly faster

Also, the PR is "backwards compatible". It searches for TXT files first, if it does not find them it uses the BIN files.

If you have any comments / questions, please let me know.

Bests,
Sebastian


cdcseacave:
Thank you @SBCV , it looks good to me. Pls point it to develop branch instead of master. I tried it myself but I get all sorts of problems.

SBCV:
See #538

cdcseacave:
thx, I'll close this one

PR #562, opened 2020-05-12
Include BUNDLE_EXT when importing VSFM scene

Eberty:
Once [file loader supports .nvm format and bundler format](https://github.com/cdcseacave/openMVS/blob/063e5cd7f1139aceaf882f9f6ae3f0de26c59138/apps/InterfaceVisualSFM/Util.h#L33), I just added an option to call [LoadBundlerOut()](https://github.com/cdcseacave/openMVS/blob/063e5cd7f1139aceaf882f9f6ae3f0de26c59138/apps/InterfaceVisualSFM/Util.h#L184) making possible convert "bundle files" (.out + .txt) into mvs project.

See: https://github.com/cdcseacave/openMVS/blob/063e5cd7f1139aceaf882f9f6ae3f0de26c59138/apps/InterfaceVisualSFM/Util.h#L414

### DIFF
--- a/apps/InterfaceVisualSFM/InterfaceVisualSFM.cpp
+++ b/apps/InterfaceVisualSFM/InterfaceVisualSFM.cpp
@@ -43,6 +43,7 @@
 #define MVS_EXT _T(".mvs")
 #define VSFM_EXT _T(".nvm")
 #define CMPMVS_EXT _T(".lst")
+#define BUNDLE_EXT _T(".out")
 
 
 // S T R U C T S ///////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


cdcseacave:
great, thx
pls point it to develop for testing first

cdcseacave:
I did some tests, and while this works, it expects only OUT projects that were already undistorted, not directly saved by VSFM, so this should be used only if you really know what are you doing, otherwise the OpenMVS results will be very poor. Plus it removes the image paths.

So pls use this patch instead, and apply it directly to develop

Eberty:
I will close this pull request and apply your suggestions into develop. Thanks.

Just a question. Is it possible to convert a project already undistorted by OpenMVS to bundler?

PR #612, opened 2020-09-09
pp.BoundaryWeight -> pp.BoundaryQuadricWeight

jiapei100:
**TriEdgeCollapseQuadricParameter** has **NO** variable **BoundaryWeight**, but **BoundaryQuadricWeight** .


cdcseacave:
as you can see the proposed change does not build; either use the VCG library I forked, or propose a change that is back compatible

PR #625, opened 2020-11-02
Add support for unsigned int 16bit and 32bit float TIFFs

pierotofy:
Hello :hand: !

I've added support for processing 16bit TIFFs (unsigned int) and 32bit floating precision TIFFs to `DensifyPointCloud`. These formats are common if you process images from multispectral cameras such as Parrot's Sequoia or Micasense cameras (popular with drones).

image

I've simplified the logic of handling RGBA images by using libtiff's built-in `TIFFReadRGBAImageOriented` for 8bit images. Ran some tests and everything seems to work OK.

Please let me know if changes are needed? :pray: 

I can provide some test data if needed for testing.


libs/IO/Image.h
```diff
@@ -107,6 +109,49 @@ class IO_API CImage
 	#endif
 
 protected:
+    template<typename T>
+    static void findMinMaxPercentile(void *data, Size size, T* min, T* max){
```
cdcseacave, line 113: pls use & or simply return a std::pair
and use the same formatting as used in the rest of the library

libs/IO/Image.h
```diff
@@ -107,6 +109,49 @@ class IO_API CImage
...
+        T *p = reinterpret_cast<T *>(data);
+        T aMin = p[0];
+        T aMax = p[0];
+
+        for (Size i = 1; i < size; i++){
+            if (p[i] > aMax) aMax = p[i];
+            if (p[i] < aMin) aMin = p[i];
+        }
+
+        double range = static_cast<double>(aMax - aMin);
```
cdcseacave, line 129: float everywhere here should be more than enough

libs/IO/Image.h
```diff
@@ -107,6 +109,49 @@ class IO_API CImage
 	#endif
 
 protected:
+    template<typename T>
```
cdcseacave, line 112: this is a useful function in general, pls move it in Util.h and call it ComputePrecentileMinMax

cdcseacave (changes requested):

cdcseacave:
Thank you for this contribution. To resume the support you added for more accurate TIFF images is:
- import 16 and 32 bit TIFF and convert them to 8 bit to be used as normal for the rest of the library
- per image tone-mapping to optimally use the 8 bit space

Please confirm or correct me if I understood wrong.

I think this is a good fast support for this type of images, however ideally we should:
 - add support for float images throughout the entire library
 - or at least add support for texturing module to use the original multisensor images during the final texture

libs/IO/ImageTIFF.h
```diff
@@ -21,6 +21,13 @@ namespace SEACAVE {
 
 // S T R U C T S ///////////////////////////////////////////////////
 
+// https://www.awaresystems.be/imaging/tiff/tifftags/sampleformat.html
+enum TIFF_SAMPLEFORMAT_TYPE {
+    TIFF_SAMPLEFORMAT_UINT = 1,
+    TIFF_SAMPLEFORMAT_INT = 2,
+    TIFF_SAMPLEFORMAT_IEEEFP = 3
+};
```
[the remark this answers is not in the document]
pierotofy: I moved it to ImageTIFF.ccp, it's not needed in the header. :+1:

libs/IO/Image.h
pierotofy: Done :+1:

libs/IO/Image.h
pierotofy: You're right. :+1:  I've changed it to float.

libs/IO/Image.h
pierotofy: Done :+1:

pierotofy:
> Please confirm or correct me if I understood wrong.

You are correct!

>  ideally we should:
> add support for float images throughout the entire library
> or at least add support for texturing module to use the original multisensor images during the final texture

Agreed, it would be a good future addition. I've been using mvs-texturing for texturing (the implementation of Nils Moehrle at https://github.com/nmoehrle/mvs-texturing), is there any significant difference in performance/quality between your implementation and his? I might be interested in contributing support for float images to the texturing if it can improve the quality.

libs/Common/Util.h
```diff
@@ -823,6 +823,50 @@ class GENERAL_API Util
 			lastMsgLen = msgLen;
 		}
 	};
+
+	template<typename T>
+	static std::pair<T,T> ComputePercentileMinMax(void *data, size_t size){
```
cdcseacave, line 828: I might be wrong, but `void*` shouldn't be `const T*` ?

libs/Common/Util.h
```diff
@@ -823,6 +823,50 @@ class GENERAL_API Util
...
+		T *p = reinterpret_cast<T *>(data);
+		T aMin = p[0];
+		T aMax = p[0];
+
+		for (size_t i = 1; i < size; i++) {
+			if (p[i] > aMax) aMax = p[i];
+			if (p[i] < aMin) aMin = p[i];
+		}
+
+		float range = static_cast<float>(aMax - aMin);
```
cdcseacave, line 842: const

libs/Common/Util.h
```diff
@@ -823,6 +823,50 @@ class GENERAL_API Util
...
+
+		float closestMinP = 9999.0;
+		float closestMaxP = 9999.0;
+
+		T min = 0;
+		T max = 0;
+		
+		// Get min/max values at the 10th and 90th percentile
+		for (size_t i = 0; i < size; i++) {
+			float percentile = (static_cast<float>(p[i]) - static_cast<float>(aMin)) / range;
```
cdcseacave, line 854: const
const
const

libs/Common/Util.h
```diff
@@ -823,6 +823,50 @@ class GENERAL_API Util
...
+			float minP = abs(percentile - 0.1);
```
cdcseacave, line 855: 0.1f
0.9f

libs/Common/Util.h
```diff
@@ -823,6 +823,50 @@ class GENERAL_API Util
...
+		for (size_t i = 1; i < size; i++) {
+			if (p[i] > aMax) aMax = p[i];
+			if (p[i] < aMin) aMin = p[i];
+		}
+
+		float range = static_cast<float>(aMax - aMin);
+		if (range == 0) 
+			return std::make_pair(aMin, aMax);
+
+		float closestMinP = 9999.0;
```
cdcseacave, line 846: pls add `f`, it creates warnings on Windows

libs/IO/Image.cpp
```diff
@@ -409,6 +413,47 @@ bool CImage::FilterFormat(void* pDst, PIXELFORMAT formatDst, Size strideDst, con
 				((uint8_t*)pDst)[2] = ((uint8_t*)pSrc)[1];
 			}
 			return true;
+
+		case PF_GRAYU16:{
+			// from PF_GRAYU16 to PF_R8G8B8
+
```
cdcseacave, line 419: no space lines

libs/IO/Image.h
```diff
@@ -32,6 +32,8 @@ typedef enum PIXELFORMAT_TYPE {
 	// gray
 	PF_A8,
 	PF_GRAY8,
+    PF_GRAYU16, // unsigned 16
```
cdcseacave, line 35: format (tab I think instead of spacexs)

libs/IO/ImageTIFF.cpp
```diff
@@ -413,22 +412,37 @@ HRESULT CImageTIFF::ReadHeader()
 			((photometric != 2 && photometric != 1) ||
 				(ncn != 1 && ncn != 3 && ncn != 4)))
 			bpp = 8;
-		switch (bpp) {
+
+		bool implemented = true;
+
+		switch (bpp){
 		case 8:
+			if (ncn >= 3){
```
cdcseacave, line 420: pls use the same format as the rest of OpenMVS code

cdcseacave (approved):
LGTM, mostly style issues, I know it is annoying, but we need to keep the unity of the library

libs/IO/ImageTIFF.cpp
pierotofy: I think you mean having a space between `(` and `{` right? (I added that).

pierotofy:
Hopefully I got all the things that needed to be changed. I understand the desire to keep a uniform style.

cdcseacave:
ready to merge, but pls target develop, not master

PR #650, opened 2021-02-23
Build libCommon/IO/Math as shared if BUILD_SHARED_LIBS=ON

LWisteria:
My project needs them as external libs.

Note that nothing is changed on the default setting i.e. BUILD_SHARED_LIBS=OFF.

### DIFF
--- a/libs/Common/CMakeLists.txt
+++ b/libs/Common/CMakeLists.txt
@@ -9,7 +9,7 @@ FILE(GLOB LIBRARY_FILES_H "*.h" "*.inl")
 LIST(REMOVE_ITEM LIBRARY_FILES_C ${PCH_C})
 SET(LIBRARY_FILES_C "${PCH_C};${LIBRARY_FILES_C}")
 
-cxx_library_with_type_no_pch(Common "Libs" "STATIC" "${cxx_default}"
+cxx_library_with_type_no_pch(Common "Libs" "" "${cxx_default}"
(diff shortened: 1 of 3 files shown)


cdcseacave:
your change simply removed the predefined `static` mode, but I do not see how that mode now depends on BUILD_SHARED_LIBS; it is a long time I wrote the cmake scripts, is it used somehow?

LWisteria:
@cdcseacave 

1. [cxx_library_with_type_no_pch("type")](https://github.com/cdcseacave/openMVS/blob/1de8e4bd83c37376914784d8bf47a17e80c95d7b/build/Utils.cmake#L849) is called
2. It calls [add_library(${type})](https://github.com/cdcseacave/openMVS/blob/1de8e4bd83c37376914784d8bf47a17e80c95d7b/build/Utils.cmake#L852)
3. The "type" specifies which type of libs will be created (cf. [cmake's help](https://cmake.org/cmake/help/v3.20/command/add_library.html))

> STATIC, SHARED, or MODULE may be given to specify the type of library to be created
> [snip]
>  If no type is given explicitly the type is STATIC or SHARED based on whether the current value of the variable BUILD_SHARED_LIBS is ON

I confirmed that lib(Common|IO|Math).so is created with `BUILD_SHARED_LIBS=ON` on this branch.

cdcseacave:
Thank you!

PR #656, opened 2021-03-06
Set internal linkage for all apps functions and variables

LWisteria:
I'm trying to integrate my module into openMVS apps.

I found that the global functions and variables conflict with my module, e.g., Initialize/Finalize and OPT::nMaxThreads.

The name of my (external project) is really needed to be global functions and variables.

But openMVS's is not. It can / must be local.

To solve such a problem, could you set linkages of them internal as much as possible?

Thanks.

### DIFF
--- a/apps/DensifyPointCloud/DensifyPointCloud.cpp
+++ b/apps/DensifyPointCloud/DensifyPointCloud.cpp
@@ -40,6 +40,7 @@ using namespace MVS;
 
 #define APPNAME _T("DensifyPointCloud")
 
+namespace {
 
 // S T R U C T S ///////////////////////////////////////////////////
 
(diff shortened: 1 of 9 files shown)


cdcseacave:
I do not understand the issue, can u pls elaborate? How is that you get a conflict while the 6 existing apps in OpenMVS do not have any conflict between eachother? And if you want to use OpenMVS as a library, you do not even have to compile its apps.

LWisteria:
Sorry, I couldn't explain in detail because it's confidential.

Firstly, I don't use apps as external libs.

My project customizes libMVS, some function is replaced with my functions and libMyModule.so is linked to libMVS.
And I need run openMVS's apps with the customized libMVS.

The libMyModule.so contains a function whose name conflicts with MVS's apps.

cdcseacave:
I still think there is something strange happening due to some design choices you made with the libs, but I do not see how this PR could hurt so I'll approve it

LWisteria:
> something strange

IMO I agree with you, but I can't control the design...

Leaving it aside, making linkage local as much as possible is good C++ manners ( similar to shorten the variable's lifetime).

Thank you for approving!

PR #671, opened 2021-04-26
Add --target-face-num parameter to ReconstructMesh

pierotofy:
Hello :hand:

This PR proposes the following additions/changes:

 - [x] Add a `--target-face-num` parameter to `ReconstructMesh`. Sometimes a target number of faces is known/desired and can sometimes work better for a user than using the `--decimate` parameter, which is a function of the number of existing faces.
 - [x] Fixes the logic in `ReconstructMesh` to allow the usage of `--mesh-file` without passing an openMVS project. Currently a user needs to always pass an input project (see line 177 and 230 of ReconstructMesh.cpp).

Please let me know if changes are needed? :pray:

### DIFF
--- a/apps/ReconstructMesh/ReconstructMesh.cpp
+++ b/apps/ReconstructMesh/ReconstructMesh.cpp
@@ -61,6 +61,7 @@ bool bUseFreeSpaceSupport;
 float fThicknessFactor;
 float fQualityFactor;
 float fDecimateMesh;
+int iTargetFaceNum;
 float fRemoveSpurious;
 bool bRemoveSpikes;
 unsigned nCloseHoles;
@@ -116,6 +117,7 @@ bool Initialize(size_t argc, LPCTSTR* argv)
 	boost::program_options::options_description config_clean("Clean options");
 	config_clean.add_options()
 		("decimate", boost::program_options::value(&OPT::fDecimateMesh)->default_value(1.f), "decimation factor in range (0..1] to be applied to the reconstructed surface (1 - disabled)")
+		("target-face-num", boost::program_options::value(&OPT::iTargetFaceNum)->default_value(0), "target number of faces to be applied to the reconstructed surface. (0 - disabled)")
 		("remove-spurious", boost::program_options::value(&OPT::fRemoveSpurious)->default_value(20.f), "spurious factor for removing faces with too long edges or isolated components (0 - disabled)")
 		("remove-spikes", boost::program_options::value(&OPT::bRemoveSpikes)->default_value(true), "flag controlling the removal of spike faces")
 		("close-holes", boost::program_options::value(&OPT::nCloseHoles)->default_value(30), "try to close small holes in the reconstructed surface (0 - disabled)")
@@ -172,7 +174,7 @@ bool Initialize(size_t argc, LPCTSTR* argv)
 		visible.add(generic).add(config_main).add(config_clean);
 		GET_LOG() << visible;
 	}
-	if (OPT::strInputFileName.IsEmpty())
+	if (OPT::strInputFileName.IsEmpty() && OPT::strMeshFileName.IsEmpty())
 		return false;
 	OPT::strExportType = OPT::strExportType.ToLower() == _T("obj") ? _T(".obj") : _T(".ply");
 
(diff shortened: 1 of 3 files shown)


apps/ReconstructMesh/ReconstructMesh.cpp
```diff
@@ -61,6 +61,7 @@ bool bUseFreeSpaceSupport;
 float fThicknessFactor;
 float fQualityFactor;
 float fDecimateMesh;
+int iTargetFaceNum;
```
cdcseacave, line 64: pls use unsigned instead of int as it never takes negative values

apps/ReconstructMesh/ReconstructMesh.cpp
```diff
@@ -167,12 +169,12 @@ bool Initialize(size_t argc, LPCTSTR* argv)
 	// validate input
 	Util::ensureValidPath(OPT::strInputFileName);
 	Util::ensureUnifySlash(OPT::strInputFileName);
-	if (OPT::vm.count("help") || OPT::strInputFileName.IsEmpty()) {
+	if (OPT::vm.count("help") || (OPT::strInputFileName.IsEmpty() && OPT::strMeshFileName.IsEmpty())) {
```
cdcseacave, line 172: I'd implement this differently:
pass the mesh file as the input instead as using --mesh-file param, and at the scene loading time you have an import option which if set on true, it allows to load a mesh too instead of only MVS

apps/ReconstructMesh/ReconstructMesh.cpp
```diff
@@ -305,9 +307,9 @@ int main(int argc, LPCTSTR* argv)
 		}
 
 		// clean the mesh
-		scene.mesh.Clean(OPT::fDecimateMesh, OPT::fRemoveSpurious, OPT::bRemoveSpikes, OPT::nCloseHoles, OPT::nSmoothMesh, false);
-		scene.mesh.Clean(1.f, 0.f, OPT::bRemoveSpikes, OPT::nCloseHoles, 0, false); // extra cleaning trying to close more holes
-		scene.mesh.Clean(1.f, 0.f, false, 0, 0, true); // extra cleaning to remove non-manifold problems created by closing holes
+		scene.mesh.Clean(OPT::fDecimateMesh, OPT::fRemoveSpurious, OPT::bRemoveSpikes, OPT::nCloseHoles, OPT::nSmoothMesh, false, OPT::iTargetFaceNum);
```
cdcseacave, line 310: combine iTargetFaceNum and fDecimateMesh, by converting the new param in the existing one by computing the desired ratio

cdcseacave (changes requested):

cdcseacave:
Good idea, but pls address the small suggestions I found.

apps/ReconstructMesh/ReconstructMesh.cpp
pierotofy: Hey @cdcseacave, thanks for the feedback. Could you expand more on this one?

>  at the scene loading time you have an import option which if set on true

Is this an additional flag? Could one just check for the file extension of the input too? (If it's MVS, load scene, if it's OBJ/PLY, load mesh).

cdcseacave:
pls see the function definition: https://github.com/cdcseacave/openMVS/blob/281d86a0c82612839f93a795d947d02a5a3b7210/libs/MVS/Scene.h#L78

pierotofy:
> pls use unsigned instead of int as it never takes negative values

:heavy_check_mark: 

> combine iTargetFaceNum and fDecimateMesh, by converting the new param in the existing one by computing the desired ratio

:heavy_check_mark: 

> pass the mesh file as the input instead as using --mesh-file param

:x: this doesn't seem to work, as the logic in https://github.com/cdcseacave/openMVS/blob/develop/apps/ReconstructMesh/ReconstructMesh.cpp#L291-L301 overrides the loaded mesh. I left the use of the `--mesh-file` parameter instead.

cdcseacave:
maybe I misunderstood, I thought the goal is to load a mesh to clean/decimate only
if that is the case, what I suggested is what you need, and you just need to add also `scene.mesh.IsEmpty()` check here https://github.com/cdcseacave/openMVS/blob/0f2a5a08502ced420f4c7600d7e95aecb6e0eaaa/apps/ReconstructMesh/ReconstructMesh.cpp#L251

pierotofy:
> I thought the goal is to load a mesh to clean/decimate only

It is.

With this PR, I can do:

```
OpenMVS/ReconstructMesh --mesh-file mesh.ply -o clean.ply --target-face-num 10000000
```

I get that you're trying to recommend to use the `-i` flag so that we could do:

```
OpenMVS/ReconstructMesh -i mesh.ply -o clean.ply --target-face-num 10000000
```

I don't know or I'm not understanding how you want me to code this, but feel free to edit as you see fit.

cdcseacave:
I've edited the file as I see it. I did not test it, changed it directly github, but should be close to working.

pierotofy:
Thanks @cdcseacave ! I'll test sometimes tomorrow and report back. :pray:

pierotofy:
Tested, and works great! :+1:

cdcseacave:
Cool, I'll merge it then. Thank you, Piero!

PR #712, opened 2021-10-28
Fix CUDA for older cards on Linux

### LINKED ISSUE #708: Possible CUDA patch-match reconstruction artifact issues
**Describe the bug**

Running `DensifyPointCloud` on an aerial dataset using the CUDA accelerated patch-match yields some artifacts / reconstruction issues compared to the same dataset processed using the CPU code.

CPU:

image

CUDA:

image

Link to results and input OpenMVS files: https://testhub.dronedb.app/r/pierotofy/openmvs

**To Reproduce**

1. Download the input files from https://testhub.dronedb.app/r/pierotofy/openmvs
2. Run: `DensifyPointCloud "openmvs/scene.mvs"  --resolution-level 2 --max-threads 4 --number-views-fuse 2 -w "openmvs/depthmaps"` (GPU)
3. Compare the run with:  `DensifyPointCloud "openmvs/scene.mvs"  --cuda-device -1 --resolution-level 2 --max-threads 4 --number-views-fuse 2 -w "openmvs/depthmaps"` (CPU)

**Expected behavior**

Reconstruction quality should be comparable.

**Desktop (please complete the following information):**

Arch Linux / CUDA 11.2

```
(odmdev)[piero:/code] gpuomvs(+3/-2)* 130 ± nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2021 NVIDIA Corporation
Built on Sun_Feb_14_21:12:58_PST_2021
Cuda compilation tools, release 11.2, V11.2.152
Build cuda_11.2.r11.2/compiler.29618528_0
```

Btw, I got a blast seeing the speed at which depthmaps were computed using CUDA. :boom: super-excited for this feature!

Let me know if you need additional information.

pierotofy:
Fixes #708
﻿


CMakeLists.txt
```diff
@@ -101,9 +102,40 @@ if(OpenMVS_USE_CUDA)
...
+
+			set(IDX 0)
+			foreach(GPU_ARCH ${LIST_GPU_ARCH})
+				list(GET LIST_GPU_CODE ${IDX} GPU_CODE)
+				set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -gencode=arch=${GPU_ARCH},code=${GPU_CODE}")
+				MATH(EXPR IDX "${IDX}+1")
+			endforeach()
 		endif()
+
+		set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --expt-relaxed-constexpr")
```
cdcseacave, line 137: this shouldn't be inside the OpenMVS_MAX_CUDA_COMPATIBILITY if?

cdcseacave:
the new code adds these flags on my machine:
```
-D_WINDOWS -Xcompiler="/W3 /GR /EHsc" -arch=sm_35 -gencode=arch=compute_35,code=sm_35 -gencode=arch=compute_37,code=sm_37 -gencode=arch=compute_50,code=sm_50 -gencode=arch=compute_52,code=sm_52 -gencode=arch=compute_53,code=sm_53 -gencode=arch=compute_60,code=sm_60 -gencode=arch=compute_61,code=sm_61 -gencode=arch=compute_62,code=sm_62 -gencode=arch=compute_70,code=sm_70 -gencode=arch=compute_72,code=sm_72 -gencode=arch=compute_75,code=sm_75 -gencode=arch=compute_80,code=sm_80 -gencode=arch=compute_86,code=sm_86
```
Would be better to only add the latest architecture and not all of them?

CMakeLists.txt
pierotofy: Nop, it applies to all configurations.

pierotofy:
If you only add the latest architecture, the program will not run on older cards. That's the point of `OpenMVS_MAX_CUDA_COMPATIBILITY` (compile once, run on all possible cards).

cdcseacave:
good point, so in this case we should modify the default mode to add only the latest architecture; so use the new code for both, and differ only at the end when the arch are added; what do you think?

pierotofy:
Sure, I'll make a commit with the change.

cdcseacave:
if I use the default mode the arch is set to 52 for some reason, not sure why as from your code it seems 86 is supported
any idea who is why is setting arch to 52 and not something else?

pierotofy:
The default depends on the CUDA version:

```
CUDA VERSION   Min CC   Deprecated CC  Default CC  Max CC
5.5 (and prior) 1.0       N/A             1.0
6.0             1.0       1.0             1.0
6.5             1.1       1.x             2.0
7.x             2.0       N/A             2.0
8.0             2.0       2.x             2.0      6.2
9.x             3.0       N/A             3.0      7.0
10.x            3.0       N/A             3.0      7.5 (3.0 deprecated in 10.2)
11.x            3.5       3.x,5.0         5.2      8.6 (11.0:8.0, 11.1:8.6)
```

pierotofy:
Now that I think about it, are you sure you want to use always the latest arch and not the compiler's default? This might prevent the code from running on your card, unless it's the latest models.

cdcseacave:
that is, I have 11.2
do you think it is best to use the max supported arch or the default one? I have no idea

pierotofy:
Probably the default one.

cdcseacave:
ok, so let's let it as it is

pierotofy:
Thanks for reviewing the changes!

cdcseacave:
I thank you for figuring out all this, it is very complex!!!

PR #717, opened 2021-12-04
Add missing InterfaceVisualSFM

davijo:
InterfaceVisualSFM was missing from apps/CMakeLists.txt


cdcseacave:
`VisualSFM` was removed intentionally, as the computed cameras are inaccurate; pls do not use it; I highly recommend using `OpenMVG`

PR #728, opened 2021-12-20
Fixed an error with the Util function in InterfaceOpenMVG

longhuan2018:
Util::ensureDirectorySlash  Renamed Util::ensureFolderSlash
Util::getFullFileName Renamed Util::getFileFullName


cdcseacave:
thx for the fix, but that app is deprecated, should not be used, use the convertor included in OpenMVG

PR #754, opened 2022-01-17
Fix various typos

luzpaz:
Found via `codespell -q 3 -L extit,indext,ist,kinf,nd,planed,rady,te,weigth,witdh`


cdcseacave:
thank you!

PR #780, opened 2022-03-14
Fix VCGlib update

bialasjaroslaw:
Apparently recent VCGlib updates broke some code - namespace and member name.


cdcseacave:
this was fixed already in develop branch, pls use that

PR #811, opened 2022-04-19
refine: robustify computation of refinement regularization scale

adelpit:
* Use the masked depth map for the distance calculation so it will consider only the content inside the user-defined region of interest
* Use the average of the inverse depth (harmonic mean of depth) so that the metric scales with parallax rather than depth. This will also have the benefit of increase the influence of near structures

146418653-1e6c9197-61c6-47bd-8856-c72f097f14eb

### DIFF
--- a/libs/MVS/SceneRefine.cpp
+++ b/libs/MVS/SceneRefine.cpp
@@ -155,6 +155,8 @@ public:
 		const DepthMap& depthMapA, const Camera& cameraA,
 		const DepthMap& depthMapB, const Camera& cameraB,
 		const Image32F& imageB, Image32F& imageA, BitMatrix& mask);
+    static float ComputeAverageLocalInverseDepth(
+            const DepthMap& depthMap, const BitMatrix& mask);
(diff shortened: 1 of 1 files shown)


cdcseacave (commented):
Hi Andrew, these days I had time to look over this PR, and I like the idea you are suggesting. However I was wondering if instead of computing a `RegularizationScale` for the entire image, would be better (in the same spirit of the idea) to compute it per pixel (or per triangle) like:
```
pixelRegularizationScale = SQUARE(depth / cameraA.GetFocalLength())
```
Hopefully this will be not only better, but cheaper to compute as we will not need to compute `ComputeAverageLocalInverseDepth()`

Did you try this already?

cdcseacave:
Thank you for the PR. The current regularization scale formulation is documented in \docs/design/MeshRefinement.md\ (Section 1.8) based on Vu et al. (2012) to ensure scale homogenization across both CPU and CUDA pipelines. Closing this PR.

Issue #787, opened 2022-03-18
[Question] Automatically Cropping/Removing Outliers from Sparse/Dense Point Cloud


braden-dev:
I'm not sure if this is the right place to ask this question but I haven't found help elsewhere. Currently, I am using an MVG to MVS pipeline to create a 3D model from a set of 2D images. The images are taken around an object sitting on a table. Ideally, the end 3D model would consist of just the object sitting on the table, but that's unrealistic. I'm really just trying to remove as much background noise as possible. I'm sure there are some parameters in either the OpenMVG or OpenMVS functions that I'm missing that might help with this to an extent.

Here's an example of a scene that I got from the MVG/MVS pipeline where the piece in the back is far lower than the object I want outlined in blue:
![image](https://user-images.githubusercontent.com/69637496/158923923-2aa7762f-c749-4d52-9773-9fb28697286e.png)

Here is an example of the kind of result I'm looking for, where there is a lot less in the background: 
![image](https://user-images.githubusercontent.com/69637496/158924086-e401f5b7-43c8-467d-8c47-79cce4bf7cd7.png)

The thing is that I want to automate this either in the MVG/MVS pipeline or once the pipeline finishes. I apologize if this isn't really an issue that relates to this but I'm at a loss for where else to ask for help.

Let me know if there's any other information that I can provide that would give a better understanding of my problem.

cdcseacave:
there are several ways to handle this:
- you can supply a mask for each image (should have the same name with the image, but with `.mask` added)
- you can supply a region of interest (oriented bounding box) using `--import-roi-file`
- OpenMVS can be augmented with a function to try to automatically estimate the ROI (didn't have time yet to do it though)

braden-dev:
> there are several ways to handle this:
> 
> * you can supply a mask for each image (should have the same name with the image, but with `.mask` added)
> * you can supply a region of interest (oriented bounding box) using `--import-roi-file`
> * OpenMVS can be augmented with a function to try to automatically estimate the ROI (didn't have time yet to do it though)

Got it, but like you said there isn't a way to automatically try to estimate the ROI or something like that? That's the kind of thing that I would be needing for my application. I'm going to be taking several pictures around an object and I only want that object in the final reconstruction but I don't want there to be any manual pieces in between taking the pictures and creating the final mesh/point cloud.

I found a bit of hacked solution that works for my purposes but it definitely would be improved with an automatic ROI process or something similar.

3XH6R:
[...]

Issue #330, opened 2018-05-29
ROI: automatically select the region of interest


cdcseacave:
Often the object that needs to be reconstructed is much smaller than the visible scene (environment). Automatically detect the region of interest (ROI) based on clues like camera positions and average scene distance.

cdcseacave:
Closing as stale. The OpenMVS codebase has undergone extensive refactoring and modernisation on the \develop\ branch since this issue was opened. If you believe this is still relevant, please test against the latest \develop\ branch and open a new issue with reproduction steps.

Issue #329, opened 2018-05-29
Auto-cut non interesting parts


gmiguelr87:
Hi!

I have one question. I have photographed around a person (it could have been any object) and I get a lot of irrelevant surface area that can be seen in the attached image. Is there any way not to generate that irrelevant surface? Or that only the most focused part of the image is used?
example
![39767924-27c36a94-52e8-11e8-8d3f-78d5aec2c700](https://user-images.githubusercontent.com/20948392/40658978-330d1e8e-634d-11e8-9293-643ccf91c493.jpg)

cdcseacave:
This is a know missing feature, which could relatively easy be implemented (pls see #330). Unfortunately I will not have time to implement it myself too soon.

gmiguelr87:
Hello again and thanks for your answer.
Ok, when I have time I'll try to work on it. My knowledge is not too advanced, but with some help I think I could get it.
Thanks again!

Eberty:
I am doing in this way:

Using a Colmap for SFM,  converting into MSV project, generating a dense point cloud (you also can reconstruct a dense point cloud from the colmap too). When you have a dense point cloud, clear it and reconstruct the mesh using a reconstruction like poisson (i use an own code or meshlab). This mesh is already aligned to the scene, so just run this steps:

```ReconstructMesh model.mvs --mesh-file mesh_file.ply --smooth 0
TextureMesh model_mesh.mvs```

cdcseacave:
Closing as stale. The OpenMVS codebase has undergone extensive refactoring and modernisation on the \develop\ branch since this issue was opened. If you believe this is still relevant, please test against the latest \develop\ branch and open a new issue with reproduction steps.

PR #822, opened 2022-06-04
Automatic region-of-interest detection

krahets:
The common two types of scenes are unbounded scenes and wrap-around scenes, in terms of indoor and outdoor. For wrap-around scenes, the majority point-clouds are outside the ROI. There are several relevant issues [https://github.com/cdcseacave/openMVS/issues/329](https://github.com/cdcseacave/openMVS/issues/329), [https://github.com/cdcseacave/openMVS/issues/787](https://github.com/cdcseacave/openMVS/issues/787).

This work provides methods and codes for detecting ROI automatically, which is motivated by @cdcseacave [https://github.com/cdcseacave/openMVS/issues/330](https://github.com/cdcseacave/openMVS/issues/330).

### Methods

Region-of-Interest(ROI) detection is equivalent to searching the center and the radius of the scene. ROI can be easily trimmed with these parameters.

1. Calculate the average camera position $P_{avg}$ , and the average camera direction $D_{avg}$ . (use median rather than mean because of higher robustness under skewed distribution.)
2. Consider $P_{avg}$ as the initial scene center, and calculate the average distance from all cameras to $P_{avg}$ as the initial scene radius $r$ .
3. The distribution of camera directions is likely to be skewed, that is, the scene could be statistically on one side of the cameras. Therefore, the scene center $C$ is calculated with $D_{avg}$ and $P_{avg}$ . The scene radius $r$ is updated accordingly.

$$
C = P_{avg} + \tan(\arcsin(D_{avg})) \times  r
$$

4. Since the imported SfM results usually contain sparse point-clouds and the corresponding visibility information, the OBB construction methods are divided into two cases:
    1. If contains: filter the point-clouds within a reasonable depth range according to scene radius and use them to fit a more accurate AABB (as the ROI).
    2. Else: directly construct an AABB (as the ROI) based on the estimated scene center and radius.

### Tests

The test data is the 'family' scenario of the Tanks-and-Temples dataset. Full, half, and quarter wrap-around scenes are tested to verify the robustness of the method.

#### Wrap-around scenario

family_circle_cameras
cameras + point-clouds trimmed with ROI

family_circle_roi
point-clouds trimmed with ROI

family_circle_unbounded
Unbounded point-clouds

#### Half wrap-around scenario

family_oneside_cameras
cameras + point-clouds trimmed with ROI

family_oneside_roi
point-clouds trimmed with ROI

family_oneside_unbounded
Unbounded point-clouds

#### Quarter wrap-around scenario

family_sparse_cameras
cameras + point-clouds trimmed with ROI

family_sparse_roi
point-clouds trimmed with ROI

family_sparse_unbounded
unbounded point-clouds

### DIFF
--- a/apps/DensifyPointCloud/DensifyPointCloud.cpp
+++ b/apps/DensifyPointCloud/DensifyPointCloud.cpp
@@ -57,6 +57,7 @@ String strDenseConfigFileName;
 String strExportDepthMapsName;
 float fMaxSubsceneArea;
 float fSampleMesh;
+int nROIMode;
 int nFusionMode;
 int thFilterPointCloud;
 int nExportNumViews;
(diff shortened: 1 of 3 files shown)


krahets:
I re-posted a PR https://github.com/cdcseacave/openMVS/pull/823 because this wasn't pushed to the develop branch.

PR #823, opened 2022-06-05
Automatic region-of-interest detection

krahets:
The common two types of scenes are unbounded scenes and wrap-around scenes, in terms of indoor and outdoor. For wrap-around scenes, the majority point-clouds are outside the ROI. There are several relevant issues [https://github.com/cdcseacave/openMVS/issues/329](https://github.com/cdcseacave/openMVS/issues/329), [https://github.com/cdcseacave/openMVS/issues/787](https://github.com/cdcseacave/openMVS/issues/787).

This work provides methods and codes for detecting ROI automatically, which is motivated by @cdcseacave [https://github.com/cdcseacave/openMVS/issues/330](https://github.com/cdcseacave/openMVS/issues/330).

### Methods

Region-of-Interest(ROI) detection is equivalent to searching the center and the radius of the scene. ROI can be easily trimmed with these parameters.

1. Calculate the average camera position $P_{avg}$ , and the average camera direction $D_{avg}$ . (use median rather than mean because of higher robustness under skewed distribution.)
2. Consider $P_{avg}$ as the initial scene center, and calculate the average distance from all cameras to $P_{avg}$ as the initial scene radius $r$ .
3. The distribution of camera directions is likely to be skewed, that is, the scene could be statistically on one side of the cameras. Therefore, the scene center $C$ is calculated with $D_{avg}$ and $P_{avg}$ . The scene radius $r$ is updated accordingly.

$$
C = P_{avg} + \tan(\arcsin(D_{avg})) \times  r
$$

4. Since the imported SfM results usually contain sparse point-clouds and the corresponding visibility information, the OBB construction methods are divided into two cases:
    1. If contains: filter the point-clouds within a reasonable depth range according to scene radius and use them to fit a more accurate AABB (as the ROI).
    2. Else: directly construct an AABB (as the ROI) based on the estimated scene center and radius.

### Tests

The test data is the 'family' scenario of the Tanks-and-Temples dataset. Full, half, and quarter wrap-around scenes are tested to verify the robustness of the method.

#### Wrap-around scenario

family_circle_cameras
cameras + point-clouds trimmed with ROI

family_circle_roi
point-clouds trimmed with ROI

family_circle_unbounded
Unbounded point-clouds

#### Half wrap-around scenario

family_oneside_cameras
cameras + point-clouds trimmed with ROI

family_oneside_roi
point-clouds trimmed with ROI

family_oneside_unbounded
Unbounded point-clouds

#### Quarter wrap-around scenario

family_sparse_cameras
cameras + point-clouds trimmed with ROI

family_sparse_roi
point-clouds trimmed with ROI

family_sparse_unbounded
unbounded point-clouds

### DIFF
--- a/apps/DensifyPointCloud/DensifyPointCloud.cpp
+++ b/apps/DensifyPointCloud/DensifyPointCloud.cpp
@@ -137,7 +137,7 @@ bool Initialize(size_t argc, LPCTSTR* argv)
 		("fusion-mode", boost::program_options::value(&OPT::nFusionMode)->default_value(0), "depth map fusion mode (-2 - fuse disparity-maps, -1 - export disparity-maps only, 0 - depth-maps & fusion, 1 - export depth-maps only)")
 		("filter-point-cloud", boost::program_options::value(&OPT::thFilterPointCloud)->default_value(0), "filter dense point-cloud based on visibility (0 - disabled)")
 		("export-number-views", boost::program_options::value(&OPT::nExportNumViews)->default_value(0), "export points with >= number of views (0 - disabled)")
-        ("roi-mode", boost::program_options::value(&OPT::nROIMode)->default_value(-1), "set region-of-interest mode for scene trim (-1 - unbounded scene, 0 - wrap-around scene)")
+        ("roi-mode", boost::program_options::value(&OPT::nROIMode)->default_value(0), "set region-of-interest mode (-1 - unbounded scene without trim, 0 - wrap-around scene with auto trim)")
         ;
 
 	// hidden options, allowed both on command line and
--- a/apps/DensifyPointCloud/DensifyPointCloud.cpp
+++ b/apps/DensifyPointCloud/DensifyPointCloud.cpp
@@ -57,6 +57,7 @@ String strDenseConfigFileName;
 String strExportDepthMapsName;
 float fMaxSubsceneArea;
 float fSampleMesh;
+int nROIMode;
 int nFusionMode;
 int thFilterPointCloud;
 int nExportNumViews;
@@ -136,7 +137,8 @@ bool Initialize(size_t argc, LPCTSTR* argv)
 		("fusion-mode", boost::program_options::value(&OPT::nFusionMode)->default_value(0), "depth map fusion mode (-2 - fuse disparity-maps, -1 - export disparity-maps only, 0 - depth-maps & fusion, 1 - export depth-maps only)")
 		("filter-point-cloud", boost::program_options::value(&OPT::thFilterPointCloud)->default_value(0), "filter dense point-cloud based on visibility (0 - disabled)")
 		("export-number-views", boost::program_options::value(&OPT::nExportNumViews)->default_value(0), "export points with >= number of views (0 - disabled)")
-		;
+        ("roi-mode", boost::program_options::value(&OPT::nROIMode)->default_value(-1), "set region-of-interest mode for scene trim (-1 - unbounded scene, 0 - wrap-around scene)")
+        ;
 
 	// hidden options, allowed both on command line and
 	// in config file, but will not be shown to the user
@@ -281,6 +283,11 @@ int main(int argc, LPCTSTR* argv)
 	// load and estimate a dense point-cloud
 	if (!scene.Load(MAKE_PATH_SAFE(OPT::strInputFileName)))
 		return EXIT_FAILURE;
+    if (OPT::nROIMode != -1) {
+        // detect region-of-interest
+        if(!scene.DetectROI())
+            return EXIT_FAILURE;
+    }
 	if (!OPT::strExportROIFileName.empty() && scene.IsBounded()) {
 		std::ofstream fs(MAKE_PATH_SAFE(OPT::strExportROIFileName));
 		if (!fs)
--- a/apps/DensifyPointCloud/DensifyPointCloud.cpp
+++ b/apps/DensifyPointCloud/DensifyPointCloud.cpp
@@ -57,7 +57,7 @@ String strDenseConfigFileName;
 String strExportDepthMapsName;
 float fMaxSubsceneArea;
 float fSampleMesh;
-int nROIMode;
+bool bEstimateROI;
 int nFusionMode;
 int thFilterPointCloud;
 int nExportNumViews;
@@ -137,7 +137,7 @@ bool Initialize(size_t argc, LPCTSTR* argv)
 		("fusion-mode", boost::program_options::value(&OPT::nFusionMode)->default_value(0), "depth map fusion mode (-2 - fuse disparity-maps, -1 - export disparity-maps only, 0 - depth-maps & fusion, 1 - export depth-maps only)")
 		("filter-point-cloud", boost::program_options::value(&OPT::thFilterPointCloud)->default_value(0), "filter dense point-cloud based on visibility (0 - disabled)")
 		("export-number-views", boost::program_options::value(&OPT::nExportNumViews)->default_value(0), "export points with >= number of views (0 - disabled)")
-        ("roi-mode", boost::program_options::value(&OPT::nROIMode)->default_value(0), "set region-of-interest mode (-1 - unbounded scene without trim, 0 - wrap-around scene with auto trim)")
+        ("estimate-roi", boost::program_options::value(&OPT::bEstimateROI)->default_value(true), "estimate region-of-interest (0 - disabled)")
         ;
 
 	// hidden options, allowed both on command line and
@@ -283,9 +283,9 @@ int main(int argc, LPCTSTR* argv)
 	// load and estimate a dense point-cloud
 	if (!scene.Load(MAKE_PATH_SAFE(OPT::strInputFileName)))
 		return EXIT_FAILURE;
-    if (OPT::nROIMode != -1) {
-        // detect region-of-interest
-        if(!scene.DetectROI())
+    if (OPT::bEstimateROI) {
+        // estimate region-of-interest
+        if (!scene.EstimateROI())
             return EXIT_FAILURE;
     }
 	if (!OPT::strExportROIFileName.empty() && scene.IsBounded()) {
--- a/libs/MVS/Scene.cpp
+++ b/libs/MVS/Scene.cpp
@@ -1373,70 +1373,71 @@ bool Scene::ScaleImages(unsigned nMaxResolution, REAL scale, const String& folde
 	}
 	return true;
 } // ScaleImages
-
 /*----------------------------------------------------------------*/
-// detect region-of-interest based on camera positions, directions and sparse points
+
+
+// estimate region-of-interest based on camera positions, directions and sparse points
 // scale specifies the ratio of the ROI's diameter
-bool Scene::DetectROI(float scale)
+bool Scene::EstimateROI(float scale)
 {
-    typedef CLISTDEF0IDX(Camera, uint32_t) CamArr;
-    CamArr camArr;
+    CameraArr cameras;
     FOREACH(i, images) {
         const Image &imageData = images[i];
         if (!imageData.IsValid())
             continue;
-        camArr.emplace_back(imageData.camera);
+        cameras.emplace_back(imageData.camera);
     }
-    unsigned nCams = camArr.size();
-    if (nCams < 3) {
-        VERBOSE("error: not enough valid views for the ROI detection");
+    const unsigned nCameras = cameras.size();
+    if (nCameras < 3) {
+        VERBOSE("error: not enough valid views for the ROI estimation");
         return false;
     }
     // compute the camera center and the direction median
-    typedef CLISTDEF0IDX(float, IDX) Scalars;
-    Scalars x(nCams), y(nCams), z(nCams), nx(nCams), ny(nCams), nz(nCams);
-    FOREACH(i, camArr) {
-        Point3f camC(camArr[i].C);
+    FloatArr x(nCameras), y(nCameras), z(nCameras), nx(nCameras), ny(nCameras), nz(nCameras);
+    FOREACH(i, cameras) {
+        Point3f camC(cameras[i].C);
         x[i] = camC.x;
         y[i] = camC.y;
         z[i] = camC.z;
-        Point3f camDirect(camArr[i].Direction());
+        Point3f camDirect(cameras[i].Direction());
         nx[i] = camDirect.x;
         ny[i] = camDirect.y;
         nz[i] = camDirect.z;
     }
-    CMatrix camCenter(x.GetMedian(), y.GetMedian(), z.GetMedian());
+    const CMatrix camCenter(x.GetMedian(), y.GetMedian(), z.GetMedian());
     CMatrix camDirectMed(nx.GetMedian(), ny.GetMedian(), nz.GetMedian());
-    float camDirectMedLen = (float) ((const CMatrix ::EVec) camDirectMed).norm();
+    const auto camDirectMedLen = (float)norm(camDirectMed);
     camDirectMed /= camDirectMedLen;
+    #if TD_VERBOSE != TD_VERBOSE_OFF
     if (VERBOSITY_LEVEL > 2)
         VERBOSE("The camera positions median is (%f,%f,%f), directions median and norm is (%f,%f,%f), %f",
                 camCenter.x, camCenter.y, camCenter.z, camDirectMed.x, camDirectMed.y, camDirectMed.z, camDirectMedLen);
-    Scalars camDepthArr(nCams);
-    FOREACH(i, camArr) {
-        Point3f ptCam(camArr[i].R * (camCenter - camArr[i].C));
-        camDepthArr[i] = ptCam.z;
-    }
+    #endif
+    FloatArr cameraDepths(nCameras);
+    FOREACH(i, cameras)
+        cameraDepths[i] = (float)cameras[i].PointDepth(camCenter);
     // estimate scene center and radius
-    float depthMedian = camDepthArr.GetMedian();
-    float camShiftCoeff =TAN(ASIN(CLAMP(camDirectMedLen, -0.999f, 0.999f)));
-    CMatrix sceneCenter = camCenter + camShiftCoeff * depthMedian * camDirectMed;
+    const float depthMedian = cameraDepths.GetMedian();
+    const float camShiftCoeff = TAN(ASIN(CLAMP(camDirectMedLen, -0.999f, 0.999f)));
+    const CMatrix sceneCenter = camCenter + camShiftCoeff * depthMedian * camDirectMed;
     uint32_t nNegDepth = 0;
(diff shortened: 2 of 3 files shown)


cdcseacave:
cool work, and great explanation
did you try this approach on spaces like scenes where the cameras are pointing outside the circle center, and not inside as in this object like example? how it is behaving? pls post some examples

krahets:
Thank you for the reply!

### Test of the pointing-outside case

I tested the method at the ‘auditorium’ scene in the Tanks-and-Temples dataset where the cameras are pointing outside the circle center.

As shown in the figures below, the point-clouds are not trimmed. The reason is that the scene radius was estimated as a negative number ($r < 0$) for this type of camera distribution, and the `IsBounded()` function returned false, which made the program not trim the point-clouds.

palace_pointout_cameras+pcd_roi_0
cameras + ROI point-clouds
palace_pointout_cameras+pcd_roi_1
cameras

### Tests of complex scenes

I also tested complex scenes: ‘temple’, ‘palace’ in Tanks-and-Temples datasets, in which some cameras point outside the center. The results show the approach is reliable and robust in complex wrap-around scenes.

- palace

palace_cameras+pcd_roi
cameras + roi pcd
palace_roi_pcd
roi pcd
palace_unbounded_pcd
unbounded pcd

- temple

temple_cameras+pcd_roi
cameras + roi pcd
temple_roi_pcd
roi pcd
temple_unbounded_pcd
unbounded pcd

### Discussion

The two types of the scenes:

- Panorama-like shooting (cameras pointing outside the center) is often employed in unbounded, indoor scenes. It is relatively difficult to trim this type of scene (most likely not be needed).
- Warp-around shooting (cameras pointing inside the center) is the most common in MVS datasets. This PR is helpful for these scenes.

Recommended solutions:

_(Already commits)_ Added an option in `DensifyPointcloud` . It is recommended to set the `roi-mode` as `-1 - unbounded scene` for the scenes with panorama-like shootings. 

```cpp
"roi-mode", ..., "set region-of-interest mode for scene trim (-1 - unbounded scene, 0 - wrap-around scene)
```

_(To-do)_ Add codes to detect the panorama-like shooting scene and make the program not trim it.

Please let me know if you have any comments or suggestions, thanks! : )

cdcseacave:
Thank you for the extended tests, it is very helpful for a clear idea. I agree there should be a way to automatically do ROI on any type of scene, for Panorama-like to select the entire scene, for Warp-around scenes to select the ROI. I see two options:
1. we merge it as is but disabled by default
2. turned it on by default, but make sure that it does not remove important parts from Panorama-like scenes, and then merge this PR
I prefer 2, but it is up to you.
Thank you for the great work!

krahets:
Thanks, @cdcseacave !  : )

Some updates:
- Added a safety check
    - Method: Count the number of cameras facing outwards from the center. If it exceeds 1/3 of the total number of cameras, it is considered that the ROI cannot be accurately estimated, and the scene will be regarded as unbounded, that is, not be trimmed.
    - This safety check is a strong condition, which ensures that only wrap-around scenes will be trimmed.
    - The method had been tested on all the scenes mentioned above, and worked well!
- Turned the scene trim on by default.

apps/DensifyPointCloud/DensifyPointCloud.cpp
```diff
@@ -136,7 +137,8 @@ bool Initialize(size_t argc, LPCTSTR* argv)
 		("fusion-mode", boost::program_options::value(&OPT::nFusionMode)->default_value(0), "depth map fusion mode (-2 - fuse disparity-maps, -1 - export disparity-maps only, 0 - depth-maps & fusion, 1 - export depth-maps only)")
 		("filter-point-cloud", boost::program_options::value(&OPT::thFilterPointCloud)->default_value(0), "filter dense point-cloud based on visibility (0 - disabled)")
 		("export-number-views", boost::program_options::value(&OPT::nExportNumViews)->default_value(0), "export points with >= number of views (0 - disabled)")
-		;
+        ("roi-mode", boost::program_options::value(&OPT::nROIMode)->default_value(0), "set region-of-interest mode (-1 - unbounded scene without trim, 0 - wrap-around scene with auto trim)")
```
cdcseacave, line 140: pls reformulate
```
"set region-of-interest mode (-1 - unbounded scene without trim, 0 - wrap-around scene with auto trim)"
```
in something like
```
"estimate region-of-interest (0 - disabled)"
```

apps/DensifyPointCloud/DensifyPointCloud.cpp
```diff
@@ -281,6 +283,11 @@ int main(int argc, LPCTSTR* argv)
 	// load and estimate a dense point-cloud
 	if (!scene.Load(MAKE_PATH_SAFE(OPT::strInputFileName)))
 		return EXIT_FAILURE;
+    if (OPT::nROIMode != -1) {
+        // detect region-of-interest
+        if(!scene.DetectROI())
```
cdcseacave, line 288: space between if and (
pls keep the existing style

libs/MVS/Scene.cpp
```diff
@@ -1373,4 +1373,97 @@ bool Scene::ScaleImages(unsigned nMaxResolution, REAL scale, const String& folde
 	}
 	return true;
 } // ScaleImages
+
```
cdcseacave, line 1376: move /*----------------------------------------------------------------*/ up, next to the end of the previous function (like in all the other palces)

libs/MVS/Scene.cpp
```diff
@@ -1373,4 +1373,97 @@ bool Scene::ScaleImages(unsigned nMaxResolution, REAL scale, const String& folde
...
+/*----------------------------------------------------------------*/
+// detect region-of-interest based on camera positions, directions and sparse points
+// scale specifies the ratio of the ROI's diameter
+bool Scene::DetectROI(float scale)
```
cdcseacave, line 1380: pls rename to EstimateROI (I used this formulation in OpenMVS)

libs/MVS/Scene.cpp
```diff
@@ -1373,4 +1373,97 @@ bool Scene::ScaleImages(unsigned nMaxResolution, REAL scale, const String& folde
...
+{
+    typedef CLISTDEF0IDX(Camera, uint32_t) CamArr;
```
cdcseacave, line 1382: there is already a list define, pls use `CameraArr`

libs/MVS/Scene.cpp
```diff
@@ -1373,4 +1373,97 @@ bool Scene::ScaleImages(unsigned nMaxResolution, REAL scale, const String& folde
...
+    CamArr camArr;
+    FOREACH(i, images) {
+        const Image &imageData = images[i];
+        if (!imageData.IsValid())
+            continue;
+        camArr.emplace_back(imageData.camera);
+    }
+    unsigned nCams = camArr.size();
```
cdcseacave, line 1390: const

libs/MVS/Scene.cpp
```diff
@@ -1373,4 +1373,97 @@ bool Scene::ScaleImages(unsigned nMaxResolution, REAL scale, const String& folde
...
 	return true;
 } // ScaleImages
+
+/*----------------------------------------------------------------*/
+// detect region-of-interest based on camera positions, directions and sparse points
+// scale specifies the ratio of the ROI's diameter
+bool Scene::DetectROI(float scale)
+{
+    typedef CLISTDEF0IDX(Camera, uint32_t) CamArr;
+    CamArr camArr;
```
cdcseacave, line 1383: even if I define sometimes a vector with the `Arr` termination, I usually call the variable with an `s` at the end, like in English

libs/MVS/Scene.cpp
```diff
@@ -1373,4 +1373,97 @@ bool Scene::ScaleImages(unsigned nMaxResolution, REAL scale, const String& folde
...
+            continue;
+        camArr.emplace_back(imageData.camera);
+    }
+    unsigned nCams = camArr.size();
+    if (nCams < 3) {
+        VERBOSE("error: not enough valid views for the ROI detection");
+        return false;
+    }
+    // compute the camera center and the direction median
+    typedef CLISTDEF0IDX(float, IDX) Scalars;
```
cdcseacave, line 1396: There is already `FloatArr`

libs/MVS/Scene.cpp
```diff
@@ -1373,4 +1373,97 @@ bool Scene::ScaleImages(unsigned nMaxResolution, REAL scale, const String& folde
...
+        y[i] = camC.y;
+        z[i] = camC.z;
+        Point3f camDirect(camArr[i].Direction());
+        nx[i] = camDirect.x;
+        ny[i] = camDirect.y;
+        nz[i] = camDirect.z;
+    }
+    CMatrix camCenter(x.GetMedian(), y.GetMedian(), z.GetMedian());
+    CMatrix camDirectMed(nx.GetMedian(), ny.GetMedian(), nz.GetMedian());
+    float camDirectMedLen = (float) ((const CMatrix ::EVec) camDirectMed).norm();
```
cdcseacave, line 1410: just use `norm(camDirectMed)`

libs/MVS/Scene.cpp
```diff
@@ -1373,4 +1373,97 @@ bool Scene::ScaleImages(unsigned nMaxResolution, REAL scale, const String& folde
...
+    camDirectMed /= camDirectMedLen;
+    if (VERBOSITY_LEVEL > 2)
```
cdcseacave, line 1412: pls surround with
```	#if TD_VERBOSE != TD_VERBOSE_OFF
	#endif
```

libs/MVS/Scene.cpp
```diff
@@ -1373,4 +1373,97 @@ bool Scene::ScaleImages(unsigned nMaxResolution, REAL scale, const String& folde
...
+        if (ptCam.z <= 0)
+            nNegDepth++;
+    }
+    // return if ROI cannot be detected accurately
+    if (nNegDepth >= camArr.size() / 3) {
+        VERBOSE("Inaccurate ROI detection. The scene will be treated as unbounded (no ROI)");
+        return true;
+    }
+    depthMedian = camDepthArr.GetMedian();
+    if (VERBOSITY_LEVEL > 2)
```
cdcseacave, line 1437: same

libs/MVS/Scene.cpp
```diff
@@ -1373,4 +1373,97 @@ bool Scene::ScaleImages(unsigned nMaxResolution, REAL scale, const String& folde
...
+    CMatrix camCenter(x.GetMedian(), y.GetMedian(), z.GetMedian());
+    CMatrix camDirectMed(nx.GetMedian(), ny.GetMedian(), nz.GetMedian());
+    float camDirectMedLen = (float) ((const CMatrix ::EVec) camDirectMed).norm();
+    camDirectMed /= camDirectMedLen;
+    if (VERBOSITY_LEVEL > 2)
+        VERBOSE("The camera positions median is (%f,%f,%f), directions median and norm is (%f,%f,%f), %f",
+                camCenter.x, camCenter.y, camCenter.z, camDirectMed.x, camDirectMed.y, camDirectMed.z, camDirectMedLen);
+    Scalars camDepthArr(nCams);
+    FOREACH(i, camArr) {
+        Point3f ptCam(camArr[i].R * (camCenter - camArr[i].C));
```
cdcseacave, line 1417: camera has already a function that estimates the depth only (and faster): `PointDepth()`

libs/MVS/Scene.cpp
```diff
@@ -1373,4 +1373,97 @@ bool Scene::ScaleImages(unsigned nMaxResolution, REAL scale, const String& folde
...
+        camDepthArr[i] = ptCam.z;
+    }
+    // estimate scene center and radius
+    float depthMedian = camDepthArr.GetMedian();
+    float camShiftCoeff =TAN(ASIN(CLAMP(camDirectMedLen, -0.999f, 0.999f)));
+    CMatrix sceneCenter = camCenter + camShiftCoeff * depthMedian * camDirectMed;
```
cdcseacave, line 1423: pls use const wherever appropiate

libs/MVS/Scene.cpp
```diff
@@ -1373,4 +1373,97 @@ bool Scene::ScaleImages(unsigned nMaxResolution, REAL scale, const String& folde
...
+    Point3fArr ptsInROI;
+    FOREACH(i, pointcloud.points) {
+        const PointCloud::Point &point = pointcloud.points[i];
+        const PointCloud::ViewArr &views = pointcloud.pointViews[i];
+        FOREACH(j, views) {
+            const Image &imageData = images[views[j]];
+            if (!imageData.IsValid())
+                continue;
+            const Camera &camera = imageData.camera;
+            Point3f ptCam(camera.R * (CMatrix(point) - camera.C));
```
cdcseacave, line 1450: use `TransformPointW2C`

cdcseacave: in fact you want `PointDepth()` here

libs/MVS/Scene.cpp
```diff
@@ -1373,4 +1373,97 @@ bool Scene::ScaleImages(unsigned nMaxResolution, REAL scale, const String& folde
...
+                ptsInROI.emplace_back(point);
+                break;
+            }
+        }
+    }
+    AABB3f aabbROI;
+    if (!ptsInROI.empty()) {
+        aabbROI.Set(ptsInROI.begin(), ptsInROI.size());
+        VERBOSE("Set the ROI by the estimated core points");
+    }
```
cdcseacave, line 1461: no return before else

cdcseacave (changes requested):
overall LGTM, pls address the comments though, thx

libs/MVS/Scene.cpp
krahets: Please help to confirm if it is correct, thx!

```cpp
    ...
    camDirectMed /= camDirectMedLen;
    #if TD_VERBOSE != TD_VERBOSE_OFF
    if (VERBOSITY_LEVEL > 2)
        VERBOSE("The camera positions median is (%f,%f,%f), directions median and norm is (%f,%f,%f), %f",
                camCenter.x, camCenter.y, camCenter.z, camDirectMed.x, camDirectMed.y, camDirectMed.z, camDirectMedLen);
    #endif
    ...
```

libs/MVS/Scene.cpp
krahets: Please help to confirm if it is correct, thx!
```
    ...
    depthMedian = cameraDepths.GetMedian();
    #if TD_VERBOSE != TD_VERBOSE_OFF
    if (VERBOSITY_LEVEL > 2)
        VERBOSE("The estimated scene center is (%f,%f,%f), radius is %f",
                sceneCenter.x, sceneCenter.y, sceneCenter.z, depthMedian);
    #endif
    ...
```

krahets:
Thx! All the comments has been addressed (two comments to be confirmed).
Thank you for your time and patience in explaining! I have benefited greatly from it.  : )

libs/MVS/Scene.cpp
cdcseacave: this is fine, the reason for this is that verbosity can be disabled completly, and in that case VERBOSITY_LEVEL does not even exists anymore

cdcseacave (approved):

cdcseacave:
thank you for the great work, merging it now

libs/MVS/Scene.cpp
krahets: I see. Thx!

cdcseacave:
I found in the first test I did a dataset that is not working, the entire pointcloud is removed; can you pls fix the ROI estimation?
[run3.zip](https://github.com/cdcseacave/openMVS/files/8883501/run3.zip)

krahets:
Thanks for the feedback!
I am analyzing this problem (more complicated than I thought before) and will post a solution trimming all kinds of scenes correctly.

cdcseacave:
one more dataset where the ROI is cutting the real scene
[images.zip](https://github.com/cdcseacave/openMVS/files/8885678/images.zip)

krahets:
> one more dataset where the ROI is cutting the real scene [images.zip](https://github.com/cdcseacave/openMVS/files/8885678/images.zip)

Got it. The new codes is on the way. : )

bdxuyidong:
Hello, Thanks for your great work! But I can't understand the 3rd idea in your 'Method' part. I can't tell the difference of scene center and camera center in your idea. I thought they were both in the center of the ROI scene. Can you pls do a sketch to explain the formulation in the 'Method' part? Or is there any other materials about the idea? Thanks again!

krahets:
> Hello, Thanks for your great work! But I can't understand the 3rd idea in your 'Method' part. I can't tell the difference of scene center and camera center in your idea. I thought they were both in the center of the ROI scene. Can you pls do a sketch to explain the formulation in the 'Method' part? Or is there any other materials about the idea? Thanks again!

Hi! The camera center is approximately equivalent to the scene center when the distribution of camera positions is 360 orbit.
However, for the forward-face cameras, we should not use the camera center as the center of the scene, because the scene may be far away from the cameras, which is the reason for introducing the "scene center".

More details could be found at [scene.cpp](https://github.com/cdcseacave/openMVS/blob/f62b38d5d02266452bb16dbcf153963001c09f93/libs/MVS/Scene.cpp#L1587).

bdxuyidong:
> > Hello, Thanks for your great work! But I can't understand the 3rd idea in your 'Method' part. I can't tell the difference of scene center and camera center in your idea. I thought they were both in the center of the ROI scene. Can you pls do a sketch to explain the formulation in the 'Method' part? Or is there any other materials about the idea? Thanks again!
> 
> Hi! The camera center is approximately equivalent to the scene center only for the 360 orbit cameras. However, for the forward-face cameras, we should not use the camera center as the center of the scene, because the scene may be far away from the cameras, which is the reason for introducing the "scene center".
> 
> More details could be found at [scene.cpp](https://github.com/cdcseacave/openMVS/blob/f62b38d5d02266452bb16dbcf153963001c09f93/libs/MVS/Scene.cpp#L1587).
> 
> An example of forward-facing cameras:
> 
> image

Hi my friend, thanks for your reply. I have read the code, but I still can't get the point of 'camShiftCoeff', what is the meaning of that formulation? The camDirectMeanLen(< 1) is the length of a vector, What is the point of ASIN(camDirectMeanLen)? I want to know if there is any physical meaning. Is the length of the camDirectMean happened to be a SIN of some angle? Forgive my poor english, thanks again.

krahets:
Hi @bdxuyidong, the physical meaning of `camShiftCoeff` is shown in the figure below

IMG_1460

zwl995:
Hello, I can understand the physical meaning of the above explanation, so why arcsin and tan? Do these two functions have any physical meaning? In addition, I found that there is a judgment in the code that the area of interest cannot be detected when the depth of the camera in the new scene center and the scene is negative. Is this condition a little strict? Sometimes there are a small number of cameras in the scene that do not look towards the center of the scene, as shown in the following figure:
捕获

PR #835, opened 2022-06-23
Allow Masks Subsets

pierotofy:
This PR tries to improve the logic of the image masks by allowing masks subset (masks for only a subset of the input images) to be provided.

Currently OpenMVS requires that an image mask is provided for each image in a dataset (all images must have an image mask when turning on --ignore-mask-label). The program was either crashing or returning some odd-looking point clouds when providing a masks subset.

Hope this can be useful to others. :pray:

### DIFF
--- a/libs/MVS/SceneDensify.cpp
+++ b/libs/MVS/SceneDensify.cpp
@@ -671,8 +671,10 @@ bool DepthMapsData::EstimateDepthMap(IIndex idxImage, int nGeometricIter)
 		#endif
 		if (prevDepthMapSize != size) {
 			BitMatrix mask;
-			if (OPTDENSE::nIgnoreMaskLabel >= 0 && DepthEstimator::ImportIgnoreMask(*depthData.GetView().pImageData, depthData.depthMap.size(), mask, (uint16_t)OPTDENSE::nIgnoreMaskLabel))
+			if (OPTDENSE::nIgnoreMaskLabel >= 0){
+				DepthEstimator::ImportIgnoreMask(*depthData.GetView().pImageData, depthData.depthMap.size(), mask, (uint16_t)OPTDENSE::nIgnoreMaskLabel);
 				depthData.ApplyIgnoreMask(mask);
+			}
 			DepthEstimator::MapMatrix2ZigzagIdx(size, coords, mask, MAXF(64,(int)nMaxThreads*8));
 			#if 0
 			// show pixels to be processed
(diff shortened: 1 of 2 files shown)


libs/MVS/SceneDensify.cpp
```diff
@@ -671,8 +671,10 @@ bool DepthMapsData::EstimateDepthMap(IIndex idxImage, int nGeometricIter)
 		#endif
 		if (prevDepthMapSize != size) {
 			BitMatrix mask;
-			if (OPTDENSE::nIgnoreMaskLabel >= 0 && DepthEstimator::ImportIgnoreMask(*depthData.GetView().pImageData, depthData.depthMap.size(), mask, (uint16_t)OPTDENSE::nIgnoreMaskLabel))
+			if (OPTDENSE::nIgnoreMaskLabel >= 0){
```
cdcseacave, line 674: are you sure this change is needed? it doesn't seem to do anything usefull

cdcseacave (commented):
can you pls provide a test scene that crashes without this chnage?
I do not quite see how these changes help with such a crash

libs/MVS/SceneDensify.cpp
pierotofy: If I don't call `ImportIgnoreMask`, if the image does not have a mask, `mask` does not get initialized. (Which might be the cause that triggers the crash/issue).

pierotofy: Commands:

```
DensifyPointCloud /datasets/brighton_mask/opensfm/undistorted/openmvs/scene.mvs --resolution-level 2 --dense-config-file /datasets/brighton_mask/opensfm/undistorted/openmvs/Densify.ini --min-resolution 500 --max-resolution 2022 --max-threads 4 --number-views-fuse 2 --sub-resolution-levels 2 -w /datasets/brighton_mask/opensfm/undistorted/openmvs/depthmaps -v 0 --geometric-iters 0 --ignore-mask-label 0
(no crash, just incomplete point cloud)
```

image


Followed by:

```
DensifyPointCloud --filter-point-cloud -1 -i /datasets/brighton_mask/opensfm/undistorted/openmvs/scene_dense.mvs -v 0 --geometric-iters 0 --ignore-mask-label 0
Segmentation fault (core dumped)
```

Point cloud with this patch:

image


Thanks for looking into this.

cdcseacave:
I did extensive testing, while I was not able to reproduce your problem above, I did find a reason why you might encounter the issue, however the changes propose here are not treating the problem, only the effect; pls use the last commit in `develop` which should fix the root-cause (plus some other fine-tuning to depth-map estimation)

pierotofy:
Thanks @cdcseacave , will do!

pierotofy:
Can confirm the latest commit fixes the issues reported in this thread. :clinking_glasses:

PR #848, opened 2022-07-26
Add support for powerpc*

pkubaj:


cdcseacave:
thx for the contribution, but I'm afraid I have to close this PR as it was not directed to develop branch, and it has many conflicts with it; since it is a simple change, I'll add it in develop

PR #861, opened 2022-09-19
Fix image masks on GPU

pierotofy:
Hello :wave: 

This PR fixes masking when using the GPU. Currently image masks are ignored (broken?) when using CUDA.

I think I did this correctly, but please advise if changes are needed :pray:


cdcseacave:
looking good, small update as it does not need to call it inside the loop, I didn't test this, I hope you did

pierotofy:
I did and worked. Thanks for the review!

PR #867, opened 2022-09-28
common: improve MvsScalablePipeline.py to support existing modules and resuming

4CJ7T:
- Fixed typos
- Stripped suffix from scene input then append back after valid name check to grab the wanted scene file
- Removed the use of multiple modules separated by comma
- Replaced the incremental counter and exit on first error with a check for sub scene files ending with the stripped suffix
- Added resume feature by checking if the scene has already been processed with the module suffix

PR #872, opened 2022-09-30
common: improve MvsScalablePipeline.py to support resuming

4CJ7T:


cdcseacave:
why did u create a second PR on same topic? see https://github.com/cdcseacave/openMVS/pull/867...

pls explain also the changes in the PR description

PR #873, opened 2022-09-30
boost lowdepth propagation for textureles areas

itacernea:
Textureless areas get a bad score at high resolution, so we need to assign a score below the threshold if low resolution estimated a depth in that position


cdcseacave:
LGTM, but pls add some before and after point-cloud screenshots

pierotofy:
Before:

image

After:

image

Animation:

anim

cdcseacave:
@pierotofy thank you for helping with this test; this one case does not look convincing for merging it, some new points appear but others disappear (like the hole on the road); do you have more insights on this? maybe more examples?

pierotofy:
Here's another example.

Before (~200k points):

image


After (~202k points):

image

Animation:

anim

Seems to have improved the road patches a bit. This was run on GPU.

pierotofy:
Tried a CPU run also on the first dataset:

Before (~143k points):

image

After (~141k points):

image

anim

cdcseacave:
Thank you, I'll merge this!

PR #886, opened 2022-10-17
Handle nvcc compilers that don't handle --list-gpu-code

pierotofy:
This PR modifies the CMake build to check the output of `nvcc --list-gpu-code` before using its values for CUDA max compatibility, as older versions of the compiler (e.g. `10.1`) do not have this flag and cause a compilation error.


cdcseacave (approved):
LGTM

PR #890, opened 2022-10-19
common: fix ObjModel::MaterialLib::Save() when using OpenMP

humanbeingZ:


humanbeingZ:
Really sorry about this, still not right ...

PR #891, opened 2022-10-19
common: fix ObjModel::MaterialLib::Save() when using OpenMP

humanbeingZ:
With different threads writing to the same file, the output will be interleaved.


cdcseacave (approved):
LGTM, thank you!

PR #892, opened 2022-10-20
texture: make faces have same uv winding order

humanbeingZ:
The packing algorithm allows for patch rotation. However, inside `MeshTexture::GenerateTexture()`, patches are transposed rather than rotated. As a result, faces will have different uv winding order.

The proposed change will make all faces have clockwise winding order.


humanbeingZ:
I'm not sure about all faces having counter-clockwise winding order though. It seems the `EdgeFunction()` inside Util.inl assumes clockwise winding order?

humanbeingZ:
IMHO, it's more consistent if all faces are in counter-clockwise uv winding order. To achieve that, we could flip the final `textureDiffuse` image, rather than flipping the Y axis as in [this line](https://github.com/cdcseacave/openMVS/blob/develop/libs/MVS/SceneTexture.cpp#L1913).

Issue #896, opened 2022-10-24
memleak in Octree.inl:233   	   m_root.m_child = new CELL_TYPE[CELL_TYPE::numChildren];        # openMVS-2.0.1


BrentHuang:
**Describe the bug**
A clear and concise description of what the bug is. Use `Viewer` to verify/display the input/output.

==4368== 200 bytes in 1 blocks are definitely lost in loss record 27 of 40
==4368==    at 0x4C34A2F: operator new[](unsigned long) (vg_replace_malloc.c:640)
==4368==    by 0x270A97: Insert<MVS::Mesh::FacesInserter::CreateOctree(MVS::Mesh::Octree&, const MVS::Mesh&)::<lambda(SEACAVE::TOctree<SEACAVE::cList<SEACAVE::TPoint3<float>, const SEACAVE::TPoint3<float>&, 0, 8192, unsigned int>, float, 3>::IDX_TYPE, SEACAVE::TOctree<S
EACAVE::cList<SEACAVE::TPoint3<float>, const SEACAVE::TPoint3<float>&, 0, 8192, unsigned int>, float, 3>::Type)> > (Octree.inl:233)
==4368==    by 0x270A97: Insert<MVS::Mesh::FacesInserter::CreateOctree(MVS::Mesh::Octree&, const MVS::Mesh&)::<lambda(SEACAVE::TOctree<SEACAVE::cList<SEACAVE::TPoint3<float>, const SEACAVE::TPoint3<float>&, 0, 8192, unsigned int>, float, 3>::IDX_TYPE, SEACAVE::TOctree<S
EACAVE::cList<SEACAVE::TPoint3<float>, const SEACAVE::TPoint3<float>&, 0, 8192, unsigned int>, float, 3>::Type)> > (Octree.inl:251)
==4368==    by 0x270A97: CreateOctree (Mesh.h:101)
==4368==    by 0x270A97: MeshRefine::ListCameraFaces() (SceneRefine.cpp:417)
==4368==    by 0x272E74: MeshRefine::SubdivideMesh(unsigned int, float, unsigned int, unsigned int) (SceneRefine.cpp:506)
==4368==    by 0x27370E: MVS::Scene::RefineMesh(unsigned int, unsigned int, unsigned int, float, unsigned int, unsigned int, unsigned int, unsigned int, float, unsigned int, unsigned int, float, float, float, float) (SceneRefine.cpp:1305)
==4368==    by 0x159103: main (RefineMesh.cpp:251)

[...]

PR #898, opened 2022-10-28
fix(Octree): fix octree memleak

BrentHuang:
https://github.com/cdcseacave/openMVS/issues/896

### DIFF
--- a/libs/Common/Octree.inl
+++ b/libs/Common/Octree.inl
@@ -25,7 +25,11 @@ inline TOctree<ITEMARR_TYPE,TYPE,DIMS,DATA_TYPE>::CELL_TYPE::CELL_TYPE()
 template <typename ITEMARR_TYPE, typename TYPE, int DIMS, typename DATA_TYPE>
 inline TOctree<ITEMARR_TYPE,TYPE,DIMS,DATA_TYPE>::CELL_TYPE::~CELL_TYPE()
 {
-	delete[] m_child;
+    if (m_child != NULL)
+    {
+        delete[] m_child;
+        m_child = NULL;
+    }
 } // destructor
 /*----------------------------------------------------------------*/
 
@@ -33,8 +37,11 @@ inline TOctree<ITEMARR_TYPE,TYPE,DIMS,DATA_TYPE>::CELL_TYPE::~CELL_TYPE()
 template <typename ITEMARR_TYPE, typename TYPE, int DIMS, typename DATA_TYPE>
 inline void TOctree<ITEMARR_TYPE,TYPE,DIMS,DATA_TYPE>::CELL_TYPE::Release()
 {
-	delete[] m_child;
-	m_child = NULL;
+    if (m_child != NULL)
+    {
+        delete[] m_child;
+        m_child = NULL;
+    }
 } // Release
 // swap the two octrees
 template <typename ITEMARR_TYPE, typename TYPE, int DIMS, typename DATA_TYPE>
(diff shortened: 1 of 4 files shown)


libs/Common/Octree.inl
```diff
@@ -167,6 +174,8 @@ void TOctree<ITEMARR_TYPE,TYPE,DIMS,DATA_TYPE>::_Insert(CELL_TYPE& cell, const P
 	// if this child cell needs to be divided further
 	if (bForceSplit || insertData.split(size, radius)) {
 		// init node and proceed recursively
+		cell.Release();
```
cdcseacave, line 177: cell should be always empty, pls remove the release and add an assert instead:
```
ASSERT(cell.m_child == NULL);
```

libs/MVS/SceneRefine.cpp
```diff
@@ -424,6 +424,7 @@ void MeshRefine::ListCameraFaces()
 			Mesh::FacesInserter inserter(arrCameraFaces[ID]);
 			octree.Traverse(frustum, inserter);
 		}
+        octree.Release();
```
cdcseacave, line 427: none of these explicit releases are needed, the memory is released automatically by the destructor

libs/Common/Octree.inl
```diff
@@ -25,16 +25,23 @@ inline TOctree<ITEMARR_TYPE,TYPE,DIMS,DATA_TYPE>::CELL_TYPE::CELL_TYPE()
...
 } // destructor
 /*----------------------------------------------------------------*/
 
 
 template <typename ITEMARR_TYPE, typename TYPE, int DIMS, typename DATA_TYPE>
 inline void TOctree<ITEMARR_TYPE,TYPE,DIMS,DATA_TYPE>::CELL_TYPE::Release()
 {
-	delete[] m_child;
-	m_child = NULL;
+    if (m_child != NULL)
```
cdcseacave, line 40: the if is not needed, it is done automatically inside delete

libs/Common/Octree.inl
```diff
@@ -25,16 +25,23 @@ inline TOctree<ITEMARR_TYPE,TYPE,DIMS,DATA_TYPE>::CELL_TYPE::CELL_TYPE()
 template <typename ITEMARR_TYPE, typename TYPE, int DIMS, typename DATA_TYPE>
 inline TOctree<ITEMARR_TYPE,TYPE,DIMS,DATA_TYPE>::CELL_TYPE::~CELL_TYPE()
 {
-	delete[] m_child;
+    if (m_child != NULL)
```
cdcseacave, line 28: if not needed, and setting it to null neither, as the object is destructed anyway, revert

cdcseacave (changes requested):

libs/Common/Octree.inl
BrentHuang: Selection_002
Selection_003

libs/Common/Octree.inl
BrentHuang: Do different compilers behave the same way?

libs/MVS/SceneRefine.cpp
BrentHuang: If you rely on destructors, it is recommended to make the Release method private

libs/Common/Octree.inl
cdcseacave: I see, so pls fix the double allocation instead of just adding a release

libs/MVS/SceneRefine.cpp
cdcseacave: no, Release() is still useful to call manually when you want to reuse the object, but it is also called automatically at destruction

libs/Common/Octree.inl
cdcseacave: yes

libs/Common/Octree.inl
BrentHuang: valgrind reported a memory leak. I see that the code does have new twice, but I'm not sure why.

cdcseacave:
thanks for spotting this issue, but this PR is not a valid fix, I fix it in bcad1eaadff463bbb7995d19f9eeae3e818a99c2

PR #940, opened 2023-02-16
DensifyPointCloud: --mask-path to specify folder containing mask images

kenshi84:


4CJ7T:
Don't think it should be called "ignore-mask-path", as it makes it sound the path is being ignored. "ignore-mask-label" is describing what it does but I guess a mask can also be called an ignore mask.

kenshi84:
@4CJ7T 
Thanks for your feedback, I modified the patch acoordingly.

I chose that name because of the `--ignore-mask-label` arg name and function names `ApplyIgnoreMask` & `ImportIgnoreMask` in DepthMap.cpp which made me think that "ignore mask" is some special term in this context. But I agree, just simply using "mask" makes sense and should suffice.

apps/DensifyPointCloud/DensifyPointCloud.cpp
```diff
@@ -296,6 +298,19 @@ int main(int argc, LPCTSTR* argv)
...
+		if (OPTDENSE::nIgnoreMaskLabel < 0) {
+			VERBOSE("--ignore-mask-label should be >=0 when --mask-path is set");
+			return EXIT_FAILURE;
+		}
+		for (Image& image : scene.images) {
+			if (!image.maskName.empty()) {
+				VERBOSE("Image %s has non-empty maskName %s", image.name.c_str(), image.maskName.c_str());
+				return EXIT_FAILURE;
+			}
+			image.maskName = Util::ensureFolderSlash(OPT::strMaskPath) + Util::getFileNameExt(image.name);
```
cdcseacave, line 309: you could also add here a check that the composed mask file name actually exists

cdcseacave (changes requested):
LGTM, I will merge it after you point it to develop instead of master

kenshi84:
Rebased onto the `develop` branch, which made me aware of this patch https://github.com/cdcseacave/openMVS/commit/2874f19fbcdf3c4d7c3b98866b7d57f8129dac29 which offers a simple way to specify mask images. Given this patch, I decided to use the same file extension (`.mask.png`) to avoid chance of confusion by the user. Also added the file existence check.

apps/DensifyPointCloud/DensifyPointCloud.cpp
```diff
@@ -296,6 +298,23 @@ int main(int argc, LPCTSTR* argv)
 	// load and estimate a dense point-cloud
 	if (!scene.Load(MAKE_PATH_SAFE(OPT::strInputFileName)))
 		return EXIT_FAILURE;
+	if (!OPT::strMaskPath.empty()) {
+		if (OPTDENSE::nIgnoreMaskLabel < 0) {
```
cdcseacave, line 302: this check is probably not needed, as all you do here is setting the maskName variable, which will not be used if not requested; however can be saved and added to the MVS scene file; what do you think?

apps/DensifyPointCloud/DensifyPointCloud.cpp
```diff
@@ -296,6 +298,23 @@ int main(int argc, LPCTSTR* argv)
...
+			VERBOSE("--ignore-mask-label should be >=0 when --mask-path is set");
+			return EXIT_FAILURE;
+		}
+		for (Image& image : scene.images) {
+			if (!image.maskName.empty()) {
+				VERBOSE("Image %s has non-empty maskName %s", image.name.c_str(), image.maskName.c_str());
+				return EXIT_FAILURE;
+			}
+			image.maskName = Util::ensureFolderSlash(OPT::strMaskPath) + Util::getFileName(image.name) + ".mask.png";
```
cdcseacave, line 311: Util::ensureFolderSlash(OPT::strMaskPath) needs to be run once at the beginning only; and you should use ensureValidFolderPath instead, which does more checks

apps/DensifyPointCloud/DensifyPointCloud.cpp
```diff
@@ -296,6 +298,23 @@ int main(int argc, LPCTSTR* argv)
...
+			if (!File::access(image.maskName)) {
+				VERBOSE("Mask image %s not found in %s", image.maskName.c_str(), OPT::strMaskPath.c_str());
```
cdcseacave, line 313: do not print strMaskPath as that is included in the maskName now

cdcseacave (changes requested):

kenshi84:
@cdcseacave Thanks for your feedback, modified the patch accordingly.

PR #951, opened 2023-03-13
Fix InterfaceOpenMVG and InterfaceVisualSFM build

yanxke:
Add InterfaceOpenMVG and InterfaceVisualSFM to the default builds.  Fix broken build with old interfaces.


cdcseacave:
InterfaceOpenMVG and InterfaceVisualSFM are not buid because they are deprecated, none should be used:
OpenMVG has an exporter for OpenMVS which should be used instead
and VisualSFM is deprecated, both the app and the file format are not doing a good job for these days, use something else instead, like OpenMVG or COLMAP

if you point this PR to master I will merge only the fixes, not the CMake change

yanxke:
I've changed the target branch and removed the CMake, though this risks the having the build fail in the future.  If they are truly deprecated, then we should just remove the code and also the documentation pointing to their usage.  Right now the doc still suggests using InterfaceOpenMVG.

cdcseacave:
can you pls point me where in the documentation you see `InterfaceOpenMVG` mentioned? I did not find it

yanxke:
> documentation

Oh, I guess the outdated documentation is in OpenMVG https://openmvg.readthedocs.io/en/latest/software/MVS/OpenMVS/

cdcseacave:
OpenMVG is not my repo, pls post an issue there

PR #952, opened 2023-03-14
Simplify docker build and reduce build image

yanxke:
- Merge the two docker files into one to simplify management.
- Put build process in shell script to and remove source and object files to reduce docker image.


docker/Dockerfile
```diff
@@ -1,48 +1,15 @@
-FROM ubuntu:22.04
+ARG BASE_IMAGE=ubuntu:22.04
+
+FROM $BASE_IMAGE
 
 ARG MASTER
```
cdcseacave, line 5: master seems not reproduced, can you pass the argument to the script?

docker/Dockerfile
```diff
@@ -1,48 +1,15 @@
...
 ARG USER_ID
 ARG GROUP_ID
```
cdcseacave, line 7: is this still needed

docker/README.md
```diff
@@ -14,7 +14,11 @@
 
 You can also build the docker image from scratch based on the **Dockerfile** (perhaps with your own changes / modifications) using:
 
-        ./buildFromScratch.sh /path/where/your/SFM/results/are
+```sh
+./buildFromScratch.sh /path/where/your/SFM/results/are
+# With CUDA support:
+./buildFromScratch_CUDA.sh /path/where/your/SFM/results/are
```
cdcseacave, line 20: can you remove the buildFromScratch_CUDA.sh too and add a param to the normal script?

docker/buildInDocker.sh
```diff
@@ -0,0 +1,78 @@
...
+fi
+
+mkdir openMVS_build
+cd openMVS_build &&\
+    cmake . ../openMVS -DCMAKE_BUILD_TYPE=Release -DVCG_ROOT=/vcglib $OPENMVS_BUILD_ARG
+
+# Install OpenMVS library
+make -j4 &&\
+    make install &&\
+    cd .. && rm -rf openMVS_build openMVS vcglib
```
cdcseacave, line 73: removing the source code is fine in general I guess, but we need to make sure the install really copies everything needed: header files, libraries, and helper scripts, like the python scripts, can you pls test that?

cdcseacave (changes requested):
Interesting approach, I am not a docker expert but I imagine running a script inside instead of docker commands has some disadvantages (like reusing images maybe), though I am fine with it, so I will merge it.

But before I do, pls address my comments and change the PR base to develop branch.

cdcseacave:
also pls synchronize with @satyajitghana which also changed this files, see #954

docker/buildInDocker.sh
yanxke: Are there test suites I can run? I'm not familiar with the structure of MVS to reliably test this manually.

yanxke:
> Interesting approach, I am not a docker expert but I imagine running a script inside instead of docker commands has some disadvantages (like reusing images maybe), though I am fine with it, so I will merge it.
> 
> But before I do, pls address my comments and change the PR base to develop branch.

The disadvantage of running the script inside the Docker container is that you can't use the partially built image and resume from there if the script fails in the middle. Since we don't expect the script to fail often, and the original build script explicitly disables the build cache, this is not an issue.

docker/README.md
yanxke: Fixed

docker/Dockerfile
yanxke: Fixed by passing in parameters.

docker/Dockerfile
yanxke: Fixed

yanxke:
Changed target branch to develop.

PR #953, opened 2023-03-15
Update Dockerfile: installs glfw3 and glew

satyajitghana:
Added install for missing libraries, which prevented Viewer from being built

PR #954, opened 2023-03-15
added gui support to docker; viewer now works

satyajitghana:
Added install for missing libraries, which prevented Viewer from being built

Also added `run` scripts for both CPU and CUDA version of docker.


satyajitghana:
refer https://github.com/cdcseacave/openMVS/pull/953

satyajitghana:
image

cdcseacave:
also pls synchronize with @yanxke which also changed these files, see #952

yanxke:
> #952

I suggest applying #952 first and then this on top of that.  There are only minor differences between the CPU and GPU versions of the scripts and we should unify as much as possible for easier maintenance.

yanxke:
It doesn't look like the Dockerfile and build script as is is designed for local development or fast changes since it pulls from the remote repo's in the build process.  It looks like it's designed for a CI process where it's meant to be built from scratch every time to make sure everything builds correctly with the latest versions of the libraries.  For local development, it should be two build scripts -- one to install the dependent libraries, and another that actually builds the MVS code.  That way Docker build can cache the former.

Anyway, the GUI support in Docker looks neat.

satyajitghana:
We could make two Dockerfiles, one for development (copy source instead of pull from repo) and one for CI. I was thinking of adding vscode devcontainer support. I've been using one lately for building apps on top on openMVS. Since the development Dockerfile will have more stuff installed, it should be maintained separately. Sounds good?

cdcseacave:
Any solution is ok with me. @yanxke how about you?

yanxke:
Sounds good.

satyajitghana:
Added `.devcontainer` support as well (with gui). it would need a a few more things like CUDA support. can be done later. as users can refer the CUDA container build script to update the `Dockerfile`

satyajitghana:
I've added `.devcontainer` support as well. Although it would need a CUDA version of it as well, but for now works. I think users can get an idea of it by looking at `buildInDocker.sh` and modify `Dockerfile` accordingly.

yanxke:
Looks great!

satyajitghana:
@cdcseacave merge?

cdcseacave:
pls add a way to use docker on windows as well, for cuda and non-cuda version; having support for GUI it would be a bonus, but not necessary if you are not familiar enough with the OS

Ideally the scripts should be in python is has a lot of OS independent functionality, and could be only one version for both Windows and Linux

satyajitghana:
I know a way for display to work in Windows and Docker. It is to use WSL and vcxsrv. No other good way I've found yet. And that will need no further changes, this will work, just follow this guide: https://dev.to/darksmile92/run-gui-app-in-linux-docker-container-on-windows-host-4kde . It should work for both CUDA and non CUDA.

cdcseacave:
that is great, but it is not fully answering my request above: pls add a way to create and run a OpenMVS docker on Windows, that is `sh` scripts are not natively supported on windows; change them to python scripts (ideally) or at least add a `CMD` script for windows (no wsl) and update readme with how to use them on windows

yanxke:
I suggest merging this one as is and put the Windows Docker support as a separate feature.

cdcseacave:
I'll do that, but this unfortunate as before all these changes windows was supported, now I can not create a docker on windows

satyajitghana:
I don't have a windows machine to test, does this work?

```

@echo off

set WORKSPACE=%CD%

set CUDA_BUILD_ARGS=
set CUDA_RUNTIME_ARGS=
set CUDA_CONTAINER_SUFFIX=
set MASTER_ARGS=
set DISPLAY_ARGS=

:parse_args
if "%~1" == "" goto build_image
if "%~1" == "--cuda" (
  set "CUDA_BUILD_ARGS=--build-arg CUDA=1 --build-arg BASE_IMAGE=nvidia/cuda:11.8.0-devel-ubuntu22.04"
  set "CUDA_RUNTIME_ARGS=--gpus all -e NVIDIA_DRIVER_CAPABILITIES=compute,utility,graphics"
  set "CUDA_CONTAINER_SUFFIX=-cuda"
  shift
  goto parse_args
)
if "%~1" == "--master" (
  set "MASTER_ARGS=--build-arg MASTER=1"
  shift
  goto parse_args
)
if "%~1" == "--workspace" (
  set "WORKSPACE=%~2"
  shift
  shift
  goto parse_args
)
echo Unknown argument: %~1
exit /b 1

:build_image
echo Running with workspace: %WORKSPACE%

docker build -t="openmvs-ubuntu%CUDA_CONTAINER_SUFFIX%" --build-arg "USER_ID=%USERPROFILE:~-4%" --build-arg "GROUP_ID=%USERPROFILE:~-4%" %CUDA_BUILD_ARGS% %MASTER_ARGS% .
if errorlevel 1 exit /b 1

set "XSOCK=\\.\pipe\X11-unix"
set "XAUTH=%TEMP%\docker.xauth"
docker run %CUDA_RUNTIME_ARGS% --entrypoint bash --ipc=host --shm-size=4gb -w /work -v "%WORKSPACE%:/work" -e "DISPLAY=%COMPUTERNAME%:0.0" -v "%XAUTH%:%XAUTH%:rw" -e "XAUTHORITY=%XAUTH%" -it openmvs-ubuntu%CUDA_CONTAINER_SUFFIX%

exit /b 0

```

for the display vcxsrv or sone xserver for windows is required. If not installed it should still work but no display.

cdcseacave:
it does something, but it fails at some point, the issue looks obvious to you?
```
D:\Pro\OpenMVS\src\docker>buildFromScratch.cmd                                                                                                                                 Running with workspace: D:\Pro\OpenMVS\src\docker                                                                                                                              [+] Building 6.2s (8/8) FINISHED                                                                                                                                                => [internal] load build definition from Dockerfile                                                                                                                      0.0s  => => transferring dockerfile: 393B                                                                                                                                      0.0s  => [internal] load .dockerignore                                                                                                                                         0.0s  => => transferring context: 2B                                                                                                                                           0.0s  => [internal] load metadata for docker.io/library/ubuntu:22.04                                                                                                           2.2s  => [auth] library/ubuntu:pull token for registry-1.docker.io                                                                                                             0.0s  => [internal] load build context                                                                                                                                         0.0s  => => transferring context: 2.84kB                                                                                                                                       0.0s  => [1/3] FROM docker.io/library/ubuntu:22.04@sha256:67211c14fa74f070d27cc59d69a7fa9aeff8e28ea118ef3babc295a0428a6d21                                                     3.6s  => => resolve docker.io/library/ubuntu:22.04@sha256:67211c14fa74f070d27cc59d69a7fa9aeff8e28ea118ef3babc295a0428a6d21                                                     0.0s  => => sha256:67211c14fa74f070d27cc59d69a7fa9aeff8e28ea118ef3babc295a0428a6d21 1.13kB / 1.13kB                                                                            0.0s  => => sha256:7a57c69fe1e9d5b97c5fe649849e79f2cfc3bf11d10bbd5218b4eb61716aebe6 424B / 424B                                                                                0.0s  => => sha256:08d22c0ceb150ddeb2237c5fa3129c0183f3cc6f5eeb2e7aa4016da3ad02140a 2.30kB / 2.30kB                                                                            0.0s  => => sha256:2ab09b027e7f3a0c2e8bb1944ac46de38cebab7145f0bd6effebfe5492c818b6 29.53MB / 29.53MB                                                                          2.9s  => => extracting sha256:2ab09b027e7f3a0c2e8bb1944ac46de38cebab7145f0bd6effebfe5492c818b6                                                                                 0.6s  => [2/3] COPY buildInDocker.sh /tmp/buildInDocker.sh                                                                                                                     0.1s  => ERROR [3/3] RUN /tmp/buildInDocker.sh --cuda 0 --user_id \Dan --group_id \Dan --master 0 && rm /tmp/buildInDocker.sh                                                  0.3s ------                                                                                                                                                                          > [3/3] RUN /tmp/buildInDocker.sh --cuda 0 --user_id \Dan --group_id \Dan --master 0 && rm /tmp/buildInDocker.sh:                                                             #7 0.253 /bin/sh: 1: /tmp/buildInDocker.sh: not found                                                                                                                          ------                                                                                                                                                                         executor failed running [/bin/sh -c /tmp/buildInDocker.sh --cuda $CUDA --user_id $USER_ID --group_id $GROUP_ID --master $MASTER && rm /tmp/buildInDocker.sh]: exit code: 127     
```

satyajitghana:
Is it because the buildInDocker is not "executable" ? I mean on Linux I could fix that error you are getting with `chmod +x buildInDocker.sh` but not sure how that is done in windows
A quick fix would be to add this in docker (after the script is copied)

```
RUN chmod +x /tmp/buildInDocker.sh
```

also another thing could be to do  `RUN ./tmp/buildInDocker.sh` instead of `RUN /tmp/buildInDocker.sh`

cdcseacave:
does not help
image

yanxke:
I believe the issue is copying over shell scripts from the Windows environment to the Unix environment. The following modification to the Dockerfile will work, but it's kind of a hack.

RUN apt update && apt install dos2unix && dos2unix /tmp/buildInDocker.sh && /tmp/buildInDocker.sh --cuda $CUDA --user_id $USER_ID --group_id $GROUP_ID --master $MASTER && rm /tmp/buildInDocker.sh

Given that WSL2 is a simple install from the Windows Store, no more difficult than installing Python, and it supports CUDA and X windows natively with no additional installs like vcxsrv, that seems to be the easiest way to build and run Docker images on Windows.

yanxke:
Here's another solution: https://github.com/docker/for-win/issues/1340#issuecomment-716941493

cdcseacave:
the second solution solved it, the line ending is the problem, it breaks linux bash system, thanks
I'll create a PR soon after more testing

PR #958, opened 2023-03-16
interface: Fix the pointcloud is always cleared.

Sanheiii:
```cpp
if (!OPT::strPointsFileName.empty()) {
	if (!scene.pointcloud.Load(MAKE_PATH_SAFE(OPT::strPointsFileName)))
		return EXIT_FAILURE;
	ASSERT(!scene.pointcloud.IsValid());
	scene.pointcloud.pointViews.Resize(scene.pointcloud.points.GetSize());
}
...
const bool bAssignPoints(!scene.pointcloud.IsEmpty() && !scene.pointcloud.IsValid());
...
FOREACH(ID, scene.images) {
	...
	if (bAssignPoints)
		AssignPoints(imageData, ID, scene.pointcloud);
}
...
if (!scene.pointcloud.IsEmpty()) {
	RFOREACH(i, scene.pointcloud.points)
		if (scene.pointcloud.pointViews[i].size() < 2)
			scene.pointcloud.RemovePoint(i);
}
```
When I load a pointcloud file, the AssignPoints() will never be called, eventually all points are removed.
```cpp
const bool bAssignPoints(!scene.pointcloud.IsEmpty() && scene.pointcloud.IsValid());
I remove the not operator before scene.pointcloud.IsValid(), it works fine.

### DIFF
--- a/apps/InterfaceMetashape/InterfaceMetashape.cpp
+++ b/apps/InterfaceMetashape/InterfaceMetashape.cpp
@@ -803,7 +803,7 @@ int main(int argc, LPCTSTR* argv)
 
 	// undistort images
 	const String pathData(MAKE_PATH_FULL(WORKING_FOLDER_FULL, OPT::strOutputImageFolder));
-	const bool bAssignPoints(!scene.pointcloud.IsEmpty() && !scene.pointcloud.IsValid());
+	const bool bAssignPoints(!scene.pointcloud.IsEmpty() && scene.pointcloud.IsValid());
 	Util::Progress progress(_T("Processed images"), scene.images.size());
 	GET_LOGCONSOLE().Pause();
 	#ifdef _USE_OPENMP


cdcseacave:
thank you, there is indeed a bug in there, but that is not the right fix
pls see the last develop commit 0f83ba4 for the fix

Sanheiii:
> thank you, there is indeed a bug in there, but that is not the right fix pls see the last develop commit [0f83ba4](https://github.com/cdcseacave/openMVS/commit/0f83ba470dc8da69003bbea0f71979520ba574e9) for the fix

I pulled and ran the latest commit and I think the issue is still not fixed.
After reading the point cloud file, pointcloud.points and pointcloud.pointviews are not empty, so bAssignPoints is always false.
Only if bAssignPoints is true, the value will be assigned to pointView.size, otherwise the points will be removed.
图片
图片

cdcseacave:
can you pls send me a test dataset, the xml file plus the images?

Sanheiii:
> can you pls send me a test dataset, the xml file plus the images?

https://1drv.ms/u/s!AkBnOes28Fpdg8pRgrzX6hNSHlljeA?e=S85hgF

cdcseacave:
pls try the latest develop, it works now
however since you are using Metashape, you should export in BlocksExchange format instead, which is much better as it contains the point cloud and the views

Sanheiii:
> pls try the latest develop, it works now however since you are using Metashape, you should export in BlocksExchange format instead, which is much better as it contains the point cloud and the views
Thank you, now it works!

PR #967, opened 2023-04-04
texture: add image mask support

Sanheiii:
Masking would not be used in TextureMesh, now I implemented it. 
Masks are sometimes not binarized, so I think we should exclude all values that are not greater than nIgnoreMaskLabel.


Sanheiii:
Sorry, I found that in some cases it still throws an exception, I will try to fix it.

Sanheiii:
I fixed the above problem and tested it, and it works fine now.

libs/MVS/SceneTexture.cpp
```diff
@@ -535,6 +542,9 @@ bool MeshTexture::ListCameraFaces(FaceDataViewArr& facesDatas, float fOutlierThr
 				ASSERT((idxFace == NO_ID && depthMap(j,i) == 0) || (idxFace != NO_ID && depthMap(j,i) > 0));
 				if (idxFace == NO_ID)
 					continue;
+				if (!mask.empty() && !mask.isSet(j, i)) {
```
cdcseacave, line 545: this is not ok, there will be cases where the face partially overlaps the invalid region of the mask, so the few pixels that remain valid will assign the view to the mask even though the face contains invalid pixels; pls see PR #968 for a correct implementation; pls build on top of it; if so, pls move the mask estimation in that PR ourside under a command line paramter, to be estimated if the scene does not contain mask files.

Sanheiii: Ok, I understand the right way to implement it, I will finish it.

cdcseacave:
looking good, great work, I still wait for the #968 PR to be fixed, and I'll let you know when it is merged

EliDavis3D:
It looks like #968 is merged, does that mean this is valid now?

cdcseacave:
@Sanheiii thx for this PR; I modified it a bit though: I removed all mask processing (should be done outside OpenMVS, when the masks are produced, including merging more labels into one, and erosion, etc), and unified the mask loading and usage

PR #988, opened 2023-05-12
Compute texture patch

MikeCernea:
Add a method to calculate texture patches and added a feature in Viewer to visualize the texture for a selected face.
In Viewer open a scene with cameras, then drag and drop the obj file for textured mesh.
 - hold Ctrl+Shift and click on a face to see the texture patch in a new window
 - hold Ctrl+Alt+Shift and click on a face to save the texture patch for that face in a jpg file

### DIFF
--- a/apps/Viewer/Scene.h
+++ b/apps/Viewer/Scene.h
@@ -60,6 +60,7 @@ public:
 	Window window;
 	ImageArr images; // scene photos
 	ImageArr textures; // mesh textures
+	MVS::Mesh::FaceIdxArr face_patch_ids;
 
 	OctreePoints octPoints;
 	OctreeMesh octMesh;
@@ -97,7 +98,7 @@ public:
 
 	void Center();
 	void TogleSceneBox();
-	void CastRay(const Ray3&, int);
+	void CastRay(const Ray3&, int, int);
 protected:
 	static void* ThreadWorker(void*);
 };
--- a/libs/MVS/Mesh.cpp
+++ b/libs/MVS/Mesh.cpp
@@ -4134,6 +4134,41 @@ REAL Mesh::ComputeVolume() const
 		volume += ComputeTriangleVolume(vertices[face[0]], vertices[face[1]], vertices[face[2]]);
 	return volume;
 }
+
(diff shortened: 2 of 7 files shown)


libs/MVS/Mesh.cpp
```diff
@@ -4086,6 +4134,46 @@ REAL Mesh::ComputeVolume() const
 		volume += ComputeTriangleVolume(vertices[face[0]], vertices[face[1]], vertices[face[2]]);
 	return volume;
 }
+
+String Mesh::PlotTexturePatch(const FIndex dbgFaceId, FaceIdxArr& face_patch_ids, cv::Mat& imgOut, const bool bSaveToFile) const {
```
cdcseacave, line 4138: you probably want const face_patch_ids

apps/Viewer/Scene.h
```diff
@@ -60,6 +60,7 @@ class Scene
 	Window window;
 	ImageArr images; // scene photos
 	ImageArr textures; // mesh textures
+	MVS::Mesh::FaceIdxArr face_patch_ids;
```
cdcseacave, line 63: why placing it here as you seem to be using it only in Scene::CastRay?

libs/MVS/Mesh.cpp
```diff
@@ -398,6 +398,54 @@ void Mesh::ComputeNormalVertices()
 }
 #endif
 
+uint32_t Mesh::ComputeTexturePatchFaces(FaceIdxArr& face_patch_ids) const {
```
cdcseacave, line 401: short description of what the fucntion does and what returns

libs/MVS/Mesh.cpp
cdcseacave, line 4138: short description of what the fucntion does and what returns

cdcseacave (changes requested):

cdcseacave:
Thank you for the contribution! The fix resolving relative material library and diffuse texture paths in \libs/IO/OBJ.cpp\ has been merged into \develop\. Closing this PR.

PR #1011, opened 2023-06-19
load masks from colmap-style output directory in DensifyPointCloud

dllu:
Loads masks from an arbitrary folder.

### DIFF
--- a/apps/InterfaceCOLMAP/InterfaceCOLMAP.cpp
+++ b/apps/InterfaceCOLMAP/InterfaceCOLMAP.cpp
@@ -44,6 +44,7 @@ using namespace MVS;
 #define APPNAME _T("InterfaceCOLMAP")
 #define MVS_EXT _T(".mvs")
 #define COLMAP_IMAGES_FOLDER _T("images/")
+#define COLMAP_MASKS_FOLDER _T("masks/")
 #define COLMAP_SPARSE_FOLDER _T("sparse/")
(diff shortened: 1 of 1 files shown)


cdcseacave (changes requested):
nice, but instead of adding support for importing masks in each interface, why not use instead the `--mask-path <folder>` functionality already existing in DensifyPointcloud which does exactly this?

4CJ7T:
colmap uses `image_name.image_extension.png` and --mask-path uses `image_name.mask.png`.
colmap doesn't undistort masks but can be scripted to. I hope `MAKE_PATH_REL` does what I think it does, if it does, awesome possum, it will be the bee's knees, maybe even the cat's meow.

4CJ7T:
I think it's nice as well but since colmap doesn't natively undistort masks, the image extension can be renamed to `.mask.png` in a script to work with `--mask-path`.

cdcseacave:
if that is true (colmap does no undistort the masks as well) then this change make even less sense: even if imported in MVS they will be wrongly used; so the only way to make this happen is a script using colmap and the undistoreted scene params that undistorts the masks (as at the interface point the scene does not have this information anymore), and in the process can also make sure they follow the OpenMVS file naming

cdcseacave:
Thank you for the contribution. As noted in the review, mask support is already provided across OpenMVS applications (such as \DensifyPointCloud\ and \RefineMesh\) via the \--mask-path\ command line option. Closing this PR.

PR #1017, opened 2023-06-28
dense: detect and improve tower like structure reconstruction

cdcseacave:
Depth-map estimation for tower like structures suffer from bad neighbor views selection. This is caused by the fact that traditionally SfM stage recovers many more points in the background compared to point on the tower structure, the region of interest. Combined with the fact that the background is much farther away compared to the foreground surface, results in the neighbor selection being biased towards good background surface coverage, and poor or next to none overlap for the foreground surface. 

This in turn is caused by several factors, but two are the most important:
 1. the foreground surface covers a very low area in the total image area
 2. the foreground surface in tower like scenes usually have very poor texture resulting in even less points in the sparse point cloud

The solution proposed here to overcome this problem is to detect if the structure to be reconstructed is tower like, in which case to approximate the tower coverage in 3D and sample it uniformly into a virtual sparse point cloud, which will be used for neighbor selection instead of the original point cloud.

Buy default this mode is enabled, but the tower like structure detection is very strict in deciding if it is indeed a tower. For now only scenes with up direction on positive Z are supported, with units in meters.


MikeCernea:
Here is an example of tower-mode 3:
Before:
image
image

And after:
image
image

Issue #1014, opened 2023-06-26
Export own dmaps with original  resolution, crash in DensifuPointCloud


thomas-19:
Hi @cdcseacave 

I am currently working with my own depth images acquired from a depth sensor. In Interface.h, it's mentioned that:

// See MVS::ExportDepthDataRaw() and MVS::ImportDepthDataRaw() for usage
// Example:
// - Image-resolution at which the depth-map was estimated
// - Depth-map-resolution, for now, only the same resolution as the image is supported

You recommended matching the depth map resolution with the image resolution. For my case, both my image and depth image resolutions are 1920x1440. However, when I run the densifypointcloud with the command:

"/DensifyPointCloud -i ~/Desktop/desk_test/scan/scene.mvs -o ~/Desktop/desk_test/scan/scene_densify.mvs -w ~/Desktop/desk_test/scan/ --geometric-iters 0 --postprocess-dmaps 0"

The densifypointcloud program crashes, reporting a Segmentation fault (core dumped).

Upon tracing the bug, I found it's caused by the discrepancy between the depth map and imageData resolutions in the function:

void DepthMapsData::FuseDepthMaps(PointCloud& pointcloud, bool bEstimateColor, bool bEstimateNormal)

It appears the imageData is scaled, while the depth map remains at the original resolution.

Then, I try to debug this crash. I noticed that the default --resolution-level option for DensifyPointCloud is set to 1. When I change the --resolution-level to 0, DensifyPointCloud runs successfully.

Thus, there seems to be a conflict between the recommendation in Interface.h and the default setting of DensifyPointCloud. Should we consider modifying the depth map size to match the imageData size in the DensifyPointCloud to resolve this issue?

cdcseacave:
All you said is true, however an elegant solution is a bit more
complicated: to set the depth map at any resolution and the pipeline to
still work. If you have time to make a PR with such a change that would be
helpful.

thomas-19:
Absolutely, I'd like to contribute to this project. Along with the depth map, I'm wondering if we should also rescale the normal map and confidence map?

[...]

PR #1024, opened 2023-07-10
dense: check and resize the existed dmap file to image

thomas-19:
for issue #1014 : add size check for existed depthmap and image. If their size not match, resize the depthmap to image size.

### DIFF
--- a/libs/MVS/DepthMap.cpp
+++ b/libs/MVS/DepthMap.cpp
@@ -2107,6 +2107,28 @@ bool MVS::ImportDepthDataRaw(const String& fileName, String& imageFileName,
 } // ImportDepthDataRaw
 /*----------------------------------------------------------------*/
 
+bool MVS::GetDepthMapHeaderSize(const String& fileName, cv::Size& size)
+{
+	FILE* f = fopen(fileName, "rb");
+	if (f == NULL) {
+		DEBUG("error: opening file '%s' for reading depth-data", fileName.c_str());
+		return false;
+	}
+	// read header
+	HeaderDepthDataRaw header;
+	if (fread(&header, sizeof(HeaderDepthDataRaw), 1, f) != 1 ||
+		header.name != HeaderDepthDataRaw::HeaderDepthDataRawName() ||
+		(header.type & HeaderDepthDataRaw::HAS_DEPTH) == 0 ||
+		header.depthWidth <= 0 || header.depthHeight <= 0 ||
+		header.imageWidth < header.depthWidth || header.imageHeight < header.depthHeight)
+	{
+		DEBUG("error: invalid depth-data file '%s'", fileName.c_str());
+		return false;
+	}
+	size = cv::Size(header.depthWidth, header.depthHeight);
+	fclose(f);
+	return true;
+}
 
 // compare the estimated and ground-truth depth-maps
 void MVS::CompareDepthMaps(const DepthMap& depthMap, const DepthMap& depthMapGT, uint32_t idxImage, float threshold)
--- a/libs/MVS/SceneDensify.cpp
+++ b/libs/MVS/SceneDensify.cpp
@@ -1962,6 +1962,14 @@ void Scene::DenseReconstructionEstimate(void* pData)
 			}
 			// try to load already compute depth-map for this image
 			if (depthmapComputed && data.nFusionMode >= 0) {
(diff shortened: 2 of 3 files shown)


cdcseacave (changes requested):

libs/MVS/DepthMap.cpp
```diff
@@ -2107,6 +2107,28 @@ bool MVS::ImportDepthDataRaw(const String& fileName, String& imageFileName,
 } // ImportDepthDataRaw
 /*----------------------------------------------------------------*/
 
+bool MVS::GetDepthMapHeaderSize(const String& fileName, cv::Size& size)
```
cdcseacave, line 2110: I think this was a interesting exercise, but function `ImportDepthDataRaw()` already existing could be already used for this if you set `flags` param to 0, pls check

cdcseacave: if you want you can still let this function, but instead of duplicating the code just call `ImportDepthDataRaw` and return the only the sizes;

more useful will be to rename it to `std::optional<std::pair<cv::Size, cv::Size>> GetDepthMapHeaderSizes(const String& fileName)`

libs/MVS/SceneDensify.cpp
```diff
@@ -1962,9 +1963,17 @@ void Scene::DenseReconstructionEstimate(void* pData)
 			}
 			// try to load already compute depth-map for this image
 			if (depthmapComputed && data.nFusionMode >= 0) {
+				const std::string depthMapName(ComposeDepthFilePath(data.scene.images[idx].ID, "dmap"));
+				cv::Size depthMapSize;
+				if (GetDepthMapHeaderSize(depthMapName, depthMapSize) && depthMapSize != data.scene.images[idx].GetSize()) {
+					depthData.Load(depthMapName);
+					DepthMap& depthMap = depthData.depthMap;
+					cv::resize(depthMap, depthMap, data.scene.images[idx].GetSize(), 0, 0, cv::INTER_NEAREST);
+					depthData.Save(depthMapName);
```
cdcseacave, line 1972: it is ok if you use `depthData` to load the depth-map, but after saving it back to disk you should release it from `depthData` as well

and you can combine this with the next check that loads the depth-map, and if we that check is true we do not release it

libs/MVS/SceneDensify.cpp
```diff
@@ -1962,9 +1963,17 @@ void Scene::DenseReconstructionEstimate(void* pData)
 			}
 			// try to load already compute depth-map for this image
 			if (depthmapComputed && data.nFusionMode >= 0) {
+				const std::string depthMapName(ComposeDepthFilePath(data.scene.images[idx].ID, "dmap"));
+				cv::Size depthMapSize;
+				if (GetDepthMapHeaderSize(depthMapName, depthMapSize) && depthMapSize != data.scene.images[idx].GetSize()) {
+					depthData.Load(depthMapName);
+					DepthMap& depthMap = depthData.depthMap;
+					cv::resize(depthMap, depthMap, data.scene.images[idx].GetSize(), 0, 0, cv::INTER_NEAREST);
```
cdcseacave, line 1971: resizing only the depth-map is not enough, you should resize all images inside, ex normal-map, etc

libs/MVS/SceneDensify.cpp
```diff
@@ -1962,9 +1963,17 @@ void Scene::DenseReconstructionEstimate(void* pData)
...
+					depthData.Save(depthMapName);
+				}
 				if (OPTDENSE::nOptimize & OPTDENSE::OPTIMIZE) {
-					if (!depthData.Load(ComposeDepthFilePath(depthData.GetView().GetID(), "dmap"))) {
-						VERBOSE("error: invalid depth-map '%s'", ComposeDepthFilePath(depthData.GetView().GetID(), "dmap").c_str());
+					if (!depthData.Load(depthMapName)) {
```
cdcseacave, line 1975: to make this compatible with the above change, modify the if like:
```
if (depthData.IsEmpty() && !depthData.Load(depthMapName)) {
```

cdcseacave (changes requested):

cdcseacave:
Thank you for the PR. Issue #1014 has been cleanly resolved in \develop\ via \isCachedDmapUsable()\ in \SceneDensify.cpp\, which verifies the header resolution of cached \.dmap\ files and triggers a fresh estimation if dimensions mismatch. Closing as superseded.

PR #1026, opened 2023-07-12
Update TransformScene.cpp about "transform matrix"

nianjiuhuiyi:
Regarding the parameter "--transform-file", modify the way the file is processed after reading so that the format of the text of the document passed in can be directly a transformation matrix. It will be more intuitive.

Like the "transfer.txt" rotated 180 degrees around the x axis, the contents inside can be: 
1 0 0 0
0 -1 0 0
0 0 -1 0
0 0 0 1


cdcseacave:
Thank you!

Issue #258, opened 2017-10-12
InterfaceVisualSFM - unknown image format .jpeg


nibab:
### Specifications like the version of the project, operating system, and hardware
**Project version** 0.7
**OS** Ubuntu 14.04

### Steps to reproduce the problem
1. build openMVS
2. run from command line: ../workspace/openMVS_build/bin/InterfaceVisualSFM -i model.nvm 

../workspace/openMVS_build/bin/InterfaceVisualSFM -i model.nvm 
18:33:22 [App     ] Build date: Oct 12 2017, 01:16:50
18:33:22 [App     ] CPU: Intel(R) Xeon(R) CPU E5-2650 v2 @ 2.60GHz
18:33:22 [App     ] RAM: 14.69GB Physical Memory 0B Virtual Memory
18:33:22 [App     ] OS: Linux 3.13.0-132-generic (x86_64)
18:33:22 [App     ] SSE & AVX compatible CPU & OS detected
18:33:22 [App     ] Command line: -i model.nvm
18:33:22 [App     ] Loading cameras/points: model.nvm
18:33:22 [App     ] 29 cameras; 14885 3D points; 101319 projections
18:33:23 [App     ] CUDA device 0 initialized: GRID K520 (compute capability 3.0; memory 3.94GB)
18:33:23 [IO      ] error: unknown image format '/home/ubuntu/w2/027.jpeg'
18:33:23 [App     ] error: failed loading image header
18:33:23 [App     ] error: failed reloading image '/home/ubuntu/w2/027.jpeg'
18:33:23 [App     ] error: can not read image /home/ubuntu/w2/027.jpeg

cdcseacave:
pls make sure you compiled openCV and openMVS with the same JPEG library dependency (same version at least)

cezarbabin:
openCV and openMVS have been not compiled with jpeg9

945540968:
Have you solved the problem? I just met the same one.

cdcseacave:
Even if you solve the error, I advise everyone to try to avoid using VisualSFM, the accuracy of the reconstruction is low. Please use [OpenMVG](https://github.com/openMVG/openMVG/).

fabiopoiesi:
I have the same problem. Has anyone found a solution to this?

antonyscerri:
[...]

PR #1027, opened 2023-07-14
(Image) Support .jpeg file extension

csparker247:
Solves the issue where files with the `.jpeg` extension aren't recognized as JPEG files:

```
20:40:00 [IO      ] error: unknown image format '/some/path//mvs/undistorted_images/DSC09064.jpeg'
```

While most JPEG loading issues are caused by incompatibilities between the OpenMVG and OpenCV JPEG libs, I just encountered this issue while using a containerized build that I've been successfully using with `.jpg` files for a few months. I can only find one other reported issue (#258) that seems to have had this same problem.

Thanks to @antonyscerri for pointing out the likely cause in that issue.


csparker247:
Alright, should be gtg.

Issue #295, opened 2018-01-28
Building on Mac


imperator-maximus:
Hi, on Mac I am getting the following error:

`[ 85%] Linking CXX executable ../../bin/InterfaceVisualSFM
Undefined symbols for architecture x86_64:
  "cv::imwrite(cv::String const&, cv::_InputArray const&, std::__1::vector<int, std::__1::allocator<int> > const&)", referenced from:
      SEACAVE::TImage<SEACAVE::TPixel<unsigned char> >::Save(SEACAVE::String const&) const in InterfaceVisualSFM.cpp.o
      SEACAVE::TImage<SEACAVE::TPixel<unsigned char> >::Save(SEACAVE::String const&) const in libMVS.a(Mesh.cpp.o)
      SEACAVE::TImage<SEACAVE::TPixel<unsigned char> >::Save(SEACAVE::String const&) const in libIO.a(OBJ.cpp.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [bin/InterfaceVisualSFM] Error 1
make[1]: *** [apps/InterfaceVisualSFM/CMakeFiles/InterfaceVisualSFM.dir/all] Error 2
make: *** [all] Error 2
`

DavidLee999:
I have the same problem. Do you solve it?

csparker247:
Just hit this so documenting this for others. `cv::String` was removed in OpenCV 4, so make sure you're linking against OpenCV 2 or 3.

cdcseacave:
Closing as stale. The OpenMVS codebase has undergone extensive refactoring and modernisation on the \develop\ branch since this issue was opened. If you believe this is still relevant, please test against the latest \develop\ branch and open a new issue with reproduction steps.

PR #1028, opened 2023-07-14
Remove all usages of cv::String

csparker247:
This replaces all usages of `cv::String` (now deprecated) with `String`. This was keeping me from building against OpenCV 4+ on macOS (#295).


cdcseacave:
thank you

PR #1037, opened 2023-08-03
Added argument for maximum texture size

Paul92:
This PR adds an argument `--max-texture-size` to the `TextureMesh` app to allow limiting the maximum texture size.

### Method description

The general idea is that the `Insert` method of the packers (`MaxRectsBinPack`, `GuillotineBinPack` and `SkylinePack`), instead of simply failing if it cannot pack, now packs as much as possible. The input list of patches is modified in-place, the successfully placed patches being removed, and the method returns a list of successfully placed patches. In the calling code of texture generation, some logic has been added supports growing the texture and retrying packing if possible (for the cases where the max texture size is too big, or is default), or to start a new texture. The UV coordinates are computed as if the texture files are joined horizontally, and this type of I/O has been added to the PLY reader/writer to demonstrate the functionality.

A known limitation is that the maximum texture size cannot be smaller than the largest patch - for this, a patch splitting logic needs to be added.

A few minor changes have been brought to the code:

- the introduction of a data structure that allows storing a patch alongside its original patch id. This eliminates the need of storing explicitly the ids in a separate list.
- a bit of a naming convention, using `unplacedPatches` and `placedPatches` to differentiate between the patches that are to be placed and those which already have been placed.

## Testing

The testing was done on the Sceaux Castle dataset.

The default behavior is unchanged, generating a texture file of 4096x4096:

 textured_mesh0

If the texture is limited to 3000px, we obtain a texture which is of size 3000x6000, which is then split by the I/O into two textures of 3000x3000:

 |  

Similarly, if the texture is limited at 2200px, we obtain a texture of size 2200x6600, which is then split by the I/O into three textures of 2200x2200:

 |  | 

Attempting a maximum texture size smaller than 2200 produces the following message:

```
14:52:58 [App     ] A patch of size 2169 x 117 does not fit the texture
14:52:58 [App     ] error: "the maximum texture size chosen cannot fit a patch"
```

Information regarding the way the textures should be loaded is stored in the header of the `ply` file, which for the last example looks like:

```
format binary_little_endian 1.0
comment TextureFile textured_mesh0.png
comment TextureFile textured_mesh1.png
comment TextureFile textured_mesh2.png
element vertex 486120
```

For completeness, these are the textures produced with skyline packing (`--patch-packing-heuristic 100`) at a maximum texture size of 2200:

 |  |  | 

and guillotine packing (`--patch-packing-heuristic 300`) at a maximum texture size of 2200:

 |  | 

Just to clarify, in all the examples above what is shown is first the entire texture image, as is represented in memory with the mesh and the UV coordinates, and below is the image split into the multiple texture maps of maximum size, as is exported on disk. 

The produced texture has been inspected visually to not change the appearance of the textured mesh, confirming that the UV coordinates are computed correctly.


cdcseacave (changes requested):
Fast and clean work, that addressed the core problem, thank you!
The approach used though of splitting/merging the textures at save/load time is not following the PLY or OBJ standard (or any other mesh representation standard); a bit more work is needed to actually store a list of texture images in the mesh structure and instead of having one long image with all faces UVs in it, to make the UVs relative to each texture image, but storing an additional index list, for each face specifying the texture image index where the UV for that face is defined.
Next, PLY has a direct way of directly exporting the additional index, while for OBJ you will need to create one material for each texture.

cdcseacave:
thank you @Paul92 for this great work,
I had to move this in a new PR pointing towards develop, lets close this and continue there

Issue #921, opened 2022-12-30
Error building in MacOS


henry1593572:
**Describe the bug**
I was following the build instructions, everything was working perfectly until the building step. It stopped because of too many errors.

`error: "This header is only meant to be used on x86 and x64 architecture"`
`error: use of undeclared identifier '__builtin_ia32_emms'; did you mean '__builtin_isless'?`
`error: too few arguments to function call, expected 2, have 0`
`error: use of undeclared identifier '__builtin_ia32_vec_init_v2si'`
`error: use of undeclared identifier '__builtin_ia32_packsswb'`
`error: use of undeclared identifier '__builtin_ia32_paddb'; did you mean '__builtin_arm_addg'?`
`error: first argument of MTE builtin function must be a pointer ('__v8qi' (vector of 8 'char' values) invalid)`
`fatal error: too many errors emitted, stopping now [-ferror-limit=]`

I had no idea what was going on. 

```
kuokchonweng@guojunyingdeMacBook-Pro make % cmake --build . -j4
[  1%] Building CXX object libs/Common/CMakeFiles/Common.dir/cmake_pch_arm64.hxx.pch
In file included from <built-in>:1:
In file included from /Users/kuokchonweng/openMVS/make/libs/Common/CMakeFiles/Common.dir/cmake_pch_arm64.hxx:5:
In file included from /Users/kuokchonweng/openMVS/libs/Common/Common.h:176:
In file included from /Users/kuokchonweng/openMVS/libs/Common/Types.h:161:
/Library/Developer/CommandLineTools/usr/lib/clang/14.0.0/include/xmmintrin.h:14:2: error: "This header is only meant to be used on x86 and x64 architecture"
#error "This header is only meant to be used on x86 and x64 architecture"
 ^
In file included from <built-in>:1:
In file included from /Users/kuokchonweng/openMVS/make/libs/Common/CMakeFiles/Common.dir/cmake_pch_arm64.hxx:5:
In file included from /Users/kuokchonweng/openMVS/libs/Common/Common.h:176:
In file included from /Users/kuokchonweng/openMVS/libs/Common/Types.h:161:
In file included from /Library/Developer/CommandLineTools/usr/lib/clang/14.0.0/include/xmmintrin.h:17:
/Library/Developer/CommandLineTools/usr/lib/clang/14.0.0/include/mmintrin.h:14:2: error: "This header is only meant to be used on x86 and x64 architecture"
[...]

PR #1042, opened 2023-08-10
Only enable SSE on non ARM systems

autoantwort:
Fixes https://github.com/cdcseacave/openMVS/issues/921


cdcseacave:
thank you, pls make it pointing to develop

CMakeLists.txt
```diff
@@ -51,6 +50,9 @@ ENDIF()
 # ${OpenMVS_BINARY_DIR}.
 PROJECT(OpenMVS)
 
+cmake_policy(SET CMP0127 NEW)
+cmake_dependent_option(OpenMVS_USE_SSE "Enable SSE optimizations" ON "NOT CMAKE_SYSTEM_PROCESSOR MATCHES \"^(arm|ARM|aarch64|AARCH64).\"" OFF)
```
dg0yt, line 54: Doesn't this implicitly raise the minimum required version to CMake 3.22?
Other policies are handled with guards in lines 7 to 15.
And the feature enabled by CMP0127 simples doesn't exist before CMake 3.22,  
https://cmake.org/cmake/help/latest/policy/CMP0127.html

cdcseacave: thx for spotting this; can you modify the if to not require the new policy? if not lets update cmake version

dg0yt: It should be enough to put the condition result into a variable, and use this variable value instead of the original `ON`.

PR #1059, opened 2023-09-16
common: add Python utility for reading DMAP files

4CJ7T:
* Moved python scripts to their own folder.
* Python utility for reading dmap files.
* Example for exporting dmap file content.


cdcseacave:
great work, this is so nice, thank you

I do have two suggestions though:

1. a lot of scripts start to accumulate, can you pls create a `scripts/python` folder and move all existing py scripts there, including yours
2. pls break your new file in two, one that just reads the DMAP file and return the data dictionary, and a second one that shows how to use it; the reason is the example also imports the visualization library which is not needed to read DMAP if a user only want that

4CJ7T:
Switching from struc to numpy frombuffer.

cdcseacave (approved):
this looks perfect now, thank you!

PR #1077, opened 2023-10-27
build: enforce Eigen >= 3.4

magwe:
Current master does not build as-is on Eigen < 3.4.


cdcseacave (commented):
pls point it to develop

PR #1101, opened 2024-02-02
Python scripting: remove -o for TextureMesh

xiaoxiae:
v2.3.0 changed the meaning of the -o parameter for TextureMesh and it now exports the mesh instead of the scene (so this no longer makes sense to do).


cdcseacave (changes requested):
LGTM, but pls point this to develop

xiaoxiae:
Oof, that's not how you do that, my mistake. I'll create a new PR.

Issue #1055, opened 2023-09-08
Problems trying to use depthmaps from a LIDAR sensor for DenseReconstruction


adricostas:
Hello,

I have a software that is able to create a 3D reconstruction using images captured with a phone and their poses provided by ArKit . I'm using my own triangulation and bundle adjustment code to extract the sparse pointcloud and then OpenMVS to densify it and creating the textured mesh. The reconstruction is working perfectly, but now I'm trying to use directly the depthmaps provided by the LIDAR in order to avoid the estimation step in the OpenMVS pipeline and I'm not able to do it.

As far as I read in other reported issues, I have to create the .dmap files from the LIDAR's depthmaps. To do that, I can use `MVS::HeaderDepthDataRaw`  and take the method `MVS::ExportDepthDataRaw` as an example. So, after creating my `MVS::Scene` and before calling `DenseReconstruction` I'm doing the following:

 ```
for (int i=0; i < mvsScene->images.size(); i++)
 {
            String fileName = WORKING_FOLDER + String::FormatString("depth%04u.dmap", mvsScene->images[i].ID);
            FILE *f = fopen(fileName, "wb");
            if (f == NULL)
            {
                DEBUG("error: opening file '%s' for writing depth-data", fileName.c_str());
               // return false;
            }

            // write header
            MVS::HeaderDepthDataRaw header;
            header.name = MVS::HeaderDepthDataRaw::HeaderDepthDataRawName();
            header.type = MVS::HeaderDepthDataRaw::HAS_DEPTH;
            header.imageWidth = (uint32_t)sfmScene->frames[i].depthMap.cols;
            header.imageHeight = (uint32_t)sfmScene->frames[i].depthMap.rows;
            header.depthWidth = (uint32_t)sfmScene->frames[i].depthMap.cols;
            header.depthHeight = (uint32_t)sfmScene->frames[i].depthMap.rows;
            header.dMin = sfmScene->frames[i].minDepth;
            header.dMax = sfmScene->frames[i].maxDepth;
            fwrite(&header, sizeof(MVS::HeaderDepthDataRaw), 1, f);

            // write image file name
            const String FileName(mvsScene->images[i].name);
            const uint16_t nFileNameSize((uint16_t)FileName.length());
            fwrite(&nFileNameSize, sizeof(uint16_t), 1, f);
            fwrite(FileName.c_str(), sizeof(char), nFileNameSize, f);

            // write neighbor IDs
[...]

PR #1116, opened 2024-03-08
dense: fix new confidence values computation in FilterDepthMap

adricostas:
As reported  in [#1055](https://github.com/cdcseacave/openMVS/issues/1055), the confidence values after executing the FilterDepthMap function were not in the expected range of [0,1]. The changes presented here address that issue by ensuring the confidence values fall within the expected range.


cdcseacave:
Thank you for the PR. Depth map confidence estimation and filtering was completely overhauled and recalibrated in PR #1292 (documented in \docs/design/DepthMapConfidence.md\), removing the old confidence heuristic. Closing as superseded.

PR #1134, opened 2024-04-26
Added DMAP save

Paul92:


cdcseacave:
thx for this great contribution; I reverted file rename and unifying spacing, and add ex in 6f1a7877587f0899d3ed8feb085a0b6b9463c434

PR #1150, opened 2024-06-15
target_link_libraries is required to build executable in my case.

jiapei100:
For instance:

`target_link_libraries(${VIEWER_NAME} Common IO tinyxml2)`


cdcseacave:
`target_link_libraries(${VIEWER_NAME} Common IO tinyxml2)` like call is already added in `cxx_executable_with_flags`, for example in `apps/DensifyPointCloud/CMakeLists.txt` the library `MVS` is linked, which in its turn links to `Common` and `IO`, so no need to link to them again

if this is not working for you, something else is the problem, for ex `MVS` does not expose the `Common` library dependency

PR #1161, opened 2024-07-24
Implemented ppm reader by following the default methods

kunchun-pretia:
PPM image format is quite common and it will be great if we include it in the openmvs. This is an implementation for PPM reader for the image inputs.


cdcseacave:
Thank you for the PR. OpenCV (which OpenMVS uses for image I/O) already supports loading and saving Netpbm formats (PPM, PGM, PBM) natively out of the box. Additionally, the PR implementation contained memory safety and type issues. Closing this PR.

PR #1195, opened 2025-01-21
Viewer color source

JuSiGabriel:
Adding visualisation of confidence map on point cloud in Viewer. Different confidence maps are avaible via F1 to F6 key. 
- F1 : no color ; 
- F2 : image color ; 
- F3 : original confidence-map (ZNCC) ; 
- F4 : confidence estimated from depth-map. The estimation is the mean depth difference between the main pixel and the other pixels of a 3X3 window. The mean is computed only from the 3 lower differences. This choice minimizes edge effect. 
- F5 : combination of original confidence (w1) and confidence estimated from depth-map (w2) (0.3*w1 + 0.7*w2)
- F6 : confidence from normal variance in a 3x3 window


cdcseacave:
merged in develop, thank you

Issue #1185, opened 2024-12-07
Nvidia capable dev container


FranciscoTorrinha:
I have recently cloned this repository and decided to use the provided dev container and I noticed building with CUDA is possible, however, because the container does not use the GPU I can't do this out of the box

In order to go around this issue, the [nvidia container runtime](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html) must be installed on the host machine and the base docker image must have CUDA installed. If this is done, the project can be built using CUDA inside the dev container.

This solves the "issue" but in order for a proper implementation, the user must not be forced into having NVIDIA hardware and such the build process for the dev container should be optional.

The bottom line is, I have done this kind of thing in the past, if it is of interest to the project I could open a PR with a possible solution for this.

cdcseacave:
just running:
```sh
./buildFromScratch.sh --cuda --workspace /path/where/your/SFM/results
```
should do exactly what you want

Asuranceturix:
I have tried just that, but it does not build, I get an error in the CMake configuration phase:
`-- Looking for a CUDA compiler
-- Looking for a CUDA compiler - /usr/local/cuda/bin/nvcc
CMake Error at /usr/share/cmake-3.22/Modules/CMakeDetermineCompilerId.cmake:726 (message):
  Compiling the CUDA compiler identification source file
  "CMakeCUDACompilerId.cu" failed.

  Compiler: /usr/local/cuda/bin/nvcc

  Build flags:

  Id flags: --keep;--keep-dir;tmp;-gencode=arch=compute_,code=sm_ -v

  

  The output was:

  1

  nvcc fatal : Unsupported gpu architecture 'compute_'

  

  

Call Stack (most recent call first):
  /usr/share/cmake-3.22/Modules/CMakeDetermineCompilerId.cmake:6 (CMAKE_DETERMINE_COMPILER_ID_BUILD)
  /usr/share/cmake-3.22/Modules/CMakeDetermineCompilerId.cmake:48 (__determine_compiler_id_test)
  /usr/share/cmake-3.22/Modules/CMakeDetermineCUDACompiler.cmake:298 (CMAKE_DETERMINE_COMPILER_ID)
  CMakeLists.txt:141 (ENABLE_LANGUAGE)


-- Configuring incomplete, errors occurred!`

ljjTYJR:
specify the cuda arch: `cmake .. -DCMAKE_CUDA_ARCHITECTURES=89`

[...]

PR #1222, opened 2025-06-19
Updated Docker files to build correctly on Ubuntu 24.04

Asuranceturix:
The current instructions to build a Docker container don't work because many things have changed in the code since they were last updated. This should close issue #1185 .


cdcseacave:
thank you @Asuranceturix , great contribution

PR #1236, opened 2025-11-16
Dense: Fix out of memory edge case due to footprint explosion

pierotofy:
Hey :wave:

I've tracked down a recurring (but infrequent) issue that would cause OpenMVS to attempt to resize a depthmap to a very large size and causing: 

```
Estimated depth-maps 134 (87.58%, 3m14s, ETA 27s)...terminate called after throwing an instance of 'cv::Exception'
what():  OpenCV(4.12.0) /code/SuperBuild/src/opencv/modules/core/src/alloc.cpp:73: error: (-4:Insufficient memory) Failed to allocate 68896563262380 bytes in function 'OutOfMemoryError'
```

This seems to happen when noise from an input sparse point cloud places a point really, really close to a view. When that happens, the `footprint1` can end up being (`focal / very very small depth value`) --> `very large number`. In turn, this causes `fScaleRatio` to be set to a very large number, which pushes the `avgScale` value, and thus `neighbor.scale` to be a very large number. 

This causes the call to:

```
if (DepthData::ViewData::ScaleImage(viewTrg.image, viewTrg.image, viewTrg.scale)) // <-- viewTrg.scale could be really large
```

In `InitViews` to crash the program.

One possible solution here is to bump the depth checks from 0 to a reasonable epsilon (I've picked 0.01, because the dataset where I reproduced this had a depth value of 0.000001), or another one could be to track a median scale rather than average (which would filter out noisy scale values).

### DIFF
--- a/libs/MVS/Scene.cpp
+++ b/libs/MVS/Scene.cpp
@@ -1293,7 +1293,7 @@ bool Scene::SelectNeighborViews(uint32_t ID, IndexArr& points, unsigned nMinView
 		const PointCloud::Point& point = pointcloud.points[idx];
 		const Depth depth((float)imageData.camera.PointDepth(point));
 		ASSERT(depth > 0);
-		if (depth <= 0)
+		if (depth <= 0.01)
 			continue;
 		// store this point
 		if (views.size() >= nMinPointViews)
@@ -1312,7 +1312,7 @@ bool Scene::SelectNeighborViews(uint32_t ID, IndexArr& points, unsigned nMinView
 			const Image& imageData2 = images[view];
 			const Depth depth2((float)imageData2.camera.PointDepth(point));
 			ASSERT(depth2 > 0);
-			if (depth2 <= 0)
+			if (depth2 <= 0.01)
 				continue;
 			const Point3f V2(imageData2.camera.C - Cast<REAL>(point));
 			const float fAngle(ACOS(ComputeAngle(V1.ptr(), V2.ptr())));


fallenmi (changes requested):
The new fixed `0.01` depth cutoff makes neighbor selection depend on the scene's arbitrary world-unit scale, and it does not actually bound the footprint ratio that causes the OOM.

`Scene::Scale` uniformly multiplies camera centers and point coordinates, so both depths are multiplied by the same factor. Meanwhile, `GetFootprintImage(depth)` is `focal / depth`, which makes `fScaleRatio` invariant under that transformation. On exact head `082cc37f1597ba750a9da28e5b7e7039520f0469`, a valid pair with depths `(1, 2)` is accepted, but the identical scene scaled by `0.001` has depths `(0.001, 0.002)` and is rejected; its footprint ratio remains `2`. Conversely, the reported pathological pair `(1e-6, 1)` is filtered at its original scale, but the identical scene scaled by `1e6` has depths `(1, 1e6)`, passes both new checks, and retains an approximately `1e6` footprint ratio. The same OOM-driving `avgScale` is therefore still possible in another valid unit system.

Please make the protection scale-invariant—for example, by robustly aggregating or bounding the dimensionless `fScaleRatio`, or by deriving any depth tolerance from scene scale—and add a regression that applies a uniform similarity scale to the same cameras/points and asserts identical neighbor selection plus a bounded `neighbor.scale`.

Verification: a warning-clean C++ oracle reproduces both cases (`head: 1 -> 0` for the valid downscaled pair, and `head: 0 -> 1` for the pathological upscaled pair while its ratio stays approximately `1e6`); `git diff --check` is clean. Exact-head CI is green on Windows, Ubuntu, and macOS. I could not run `MVSPipelineTest` locally because exact-merge CMake configuration stopped at the missing OpenCV package configuration (`OpenCVConfig.cmake`).

AI disclosure: OpenAI Codex was used to inspect the exact diff and scale/neighbor-selection paths, construct and run the deterministic oracle, check the live merge ref and interaction history, and draft this review. The cited commands and outputs were run against the exact SHAs above.

cdcseacave:
Thanks for tracking this down, the diagnosis was spot on. I went with the median you suggested instead of the depth cutoff, since a fixed 0.01 depends on the scene's units (as noted in review). It's in b52a461 on [`feature/cxx17-minimum`](https://github.com/cdcseacave/openMVS/tree/feature/cxx17-minimum):
- the neighbor scale is the median of the per-point footprint ratios, so one near-camera point can't move it, whatever the unit
- `FilterNeighborViews` now clamps the scale of every view it keeps. It only drops out-of-range views while enough remain, which is how the huge scale reached `InitViews`.
- a regression test covers the outlier at scene scales 1, 1e-3 and 1e6 (the old mean gave 20001.1)

cdcseacave:
Follow-up: instead of the plain median, the scale is now a 10%-trimmed mean of the ratios (at least one dropped per end), via a new `cList::GetTrimmedMean` (769c8e5). It's just as robust to the near-camera point and stays closer to the old mean on clean scenes.

PR #1242, opened 2025-12-16
Fix x86 MSVC compilation

dg0yt:
Upstreamed from vcpkg port. Fixes
~~~
FAILED: [code=2] apps/InterfaceMetashape/CMakeFiles/InterfaceMetashape.dir/InterfaceMetashape.cpp.obj 
C:\PROGRA~1\MICROS~1\2022\ENTERP~1\VC\Tools\MSVC\1444~1.352\bin\Hostx64\x86\cl.exe   /TP  -ID:\installed\x86-windows\include -ID:\installed\x86-windows\include\eigen3 -ID:\installed\x86-windows\include\opencv4 -ID:\b\openmvs\src\v2.3.0-1d44f1f1db.clean -ID:\b\openmvs\x86-windows-rel /nologo /DWIN32 /D_WINDOWS /utf-8 /GR /EHsc /MP   /D _CRT_SECURE_NO_DEPRECATE /D _CRT_NONSTDC_NO_DEPRECATE /D _SCL_SECURE_NO_WARNINGS /Gy /bigobj /arch:SSE2 /Oi /fp:fast /Zm170 /Zc:__cplusplus  /wd4231 /wd4251 /wd4308 /wd4396 /wd4503 /wd4661 /wd4996 /MD /O2 /Oi /Gy /DNDEBUG /Z7   /GL -std:c++17 -MD  /nologo /DWIN32 /D_WINDOWS /utf-8 /GR /EHsc /MP   /D _CRT_SECURE_NO_DEPRECATE /D _CRT_NONSTDC_NO_DEPRECATE /D _SCL_SECURE_NO_WARNINGS /Gy /bigobj /arch:SSE2 /Oi /fp:fast /Zm170 /Zc:__cplusplus  /wd4231 /wd4251 /wd4308 /wd4396 /wd4503 /wd4661 /wd4996 -DWIN32 -D_WIN32 -EHsc -D_HAS_EXCEPTIONS=1 /showIncludes /Foapps\InterfaceMetashape\CMakeFiles\InterfaceMetashape.dir\InterfaceMetashape.cpp.obj /Fdapps\InterfaceMetashape\CMakeFiles\InterfaceMetashape.dir\ /FS -c D:\b\openmvs\src\v2.3.0-1d44f1f1db.clean\apps\InterfaceMetashape\InterfaceMetashape.cpp
[...]

PR #1243, opened 2025-12-16
Don't use combination of /arch:AVX /arch:SSE2 /arch:SSE

dg0yt:
Avoid MSVC x86 compiler warning:
~~~
cl : Command line warning D9025 : overriding '/arch:SSE' with '/arch:SSE2'
~~~

(Both ENABLE_SSE and ENABLE_SSE2 are enabled by default for x86.)

### DIFF
--- a/build/Utils.cmake
+++ b/build/Utils.cmake
@@ -635,17 +635,13 @@ macro(optimize_default_compiler_settings)
 		set(BUILD_EXTRA_FLAGS_RELEASE "${BUILD_EXTRA_FLAGS_RELEASE} /Zi")
 	  endif()
 
-	  if(NOT MSVC64)
-		# 64-bit MSVC compiler uses SSE/SSE2 by default
-		if(ENABLE_SSE2)
-		  set(BUILD_EXTRA_FLAGS "${BUILD_EXTRA_FLAGS} /arch:SSE2")
-		elseif(ENABLE_SSE)
-		  set(BUILD_EXTRA_FLAGS "${BUILD_EXTRA_FLAGS} /arch:SSE")
-		endif()
-	  endif()
-
+	  # 64-bit MSVC compiler uses SSE/SSE2 by default
 	  if(ENABLE_AVX)
 		set(BUILD_EXTRA_FLAGS "${BUILD_EXTRA_FLAGS} /arch:AVX")
+	  elseif(ENABLE_SSE2 AND NOT MSVC64)
+		set(BUILD_EXTRA_FLAGS "${BUILD_EXTRA_FLAGS} /arch:SSE2")
+	  elseif(ENABLE_SSE AND NOT MSVC64)
+		set(BUILD_EXTRA_FLAGS "${BUILD_EXTRA_FLAGS} /arch:SSE")
 	  endif()
 
 	  if(ENABLE_SSE OR ENABLE_SSE2 OR ENABLE_SSE3 OR ENABLE_SSE4_1)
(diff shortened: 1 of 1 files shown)


fallenmi (changes requested):
`build/Utils.cmake:645` leaves one extra `endif()` after converting the two sibling `ENABLE_SSE` / `ENABLE_SSE2` blocks into a single `if` / `elseif`. The old block has three `if` statements and three closing statements; the new block has only two `if` statements but still has all three closings. As a result, CMake cannot parse the file, so every configure fails before the intended MSVC flag selection can run.

I parsed the exact `build/Utils.cmake` from merge-base `973a37b2e9f417a80696a1d93994e636078587aa`, exact head `7fa6d6a15bc514d75d3036b61a7e556bf7b2da85`, and current GitHub merge `2512ecd017413df3f0d534a606130e62737c2770`. The base exits 0; head and merge both exit 1 with `Flow control statements are not properly nested`. Removing only the leftover line 645 from exact head restores a clean parse. A separate four-case CMake oracle confirms the intended final selection is otherwise correct: neither option emits no `/arch` flag, SSE-only emits `/arch:SSE`, SSE2-only emits `/arch:SSE2`, and both enabled emit only `/arch:SSE2`. Oracle SHA-256: `9afb5d6ee9a924d9cd30ee48dfda686d56adafb910b300b7e6182ccbe99e1ae2`.

Please remove the extra `endif()` and add at least a CMake parse/configure check for this option matrix. This upstream head currently has no GitHub check runs, commit statuses, or Actions runs, which explains why the syntax failure is not visible in the PR UI.

Disclosure: I used OpenAI Codex to inspect the exact diff and applicable repository policy, verify the live refs, interactions, workflow layers, and current merge tree, run the exact-revision CMake parser checks and option-matrix oracle, and draft this review. I independently checked the evidence and agree with the conclusion.

build/Utils.cmake
```diff
@@ -639,11 +637,11 @@ macro(optimize_default_compiler_settings)
...
 	  if(NOT MSVC64)
 		# 64-bit MSVC compiler uses SSE/SSE2 by default
-		if(ENABLE_SSE)
-		  set(BUILD_EXTRA_FLAGS "${BUILD_EXTRA_FLAGS} /arch:SSE")
-		endif()
 		if(ENABLE_SSE2)
 		  set(BUILD_EXTRA_FLAGS "${BUILD_EXTRA_FLAGS} /arch:SSE2")
+		elseif(ENABLE_SSE)
+		  set(BUILD_EXTRA_FLAGS "${BUILD_EXTRA_FLAGS} /arch:SSE")
+		endif()
```
dg0yt:
```suggestion
```

dg0yt:
Glad the error was found, but...

I cannot believe it take so much resources to review a few lines of CMake changes, here and in #1242. Similar walls of text in both issues. Is this considered useful? Is "I" you or your agent? What if the inital posts were as noisy? If every post is meant to be read through AI summaries, just let AI fix the bugs instead.

cdcseacave:
Thanks! Taken as proposed (you're co-author) in 0d55817 on [`feature/cxx17-minimum`](https://github.com/cdcseacave/openMVS/tree/feature/cxx17-minimum), headed for develop. I checked the final block in all 8 SSE/SSE2/AVX combinations on x86 and x64: at most one `/arch`, AVX > SSE2 > SSE, and only AVX on x64. The stray `endif()` from the first revision is gone.

PR #1255, opened 2026-04-06
Orthomap generation

cdcseacave:
Orthomap (true orthophoto) generation pipeline: given an already-reconstructed mesh with Z-up orientation, the pipeline produces a geometrically correct, seamless top-down image of the scene — an orthomap where every pixel maps to a real-world XY coordinate with uniform scale. It also generates a DEM.


abk9436-boop:
Great work. Is there a timeline for the completion of Step 3 and Step 4? 

I am looking forward to test it

PR #1265, opened 2026-05-05
Make pkg-config + libjxl genuinely optional in libs/IO

ShuangLiu1992:
## Summary

JPEG-XL is an *optional* image format, but `libs/IO/CMakeLists.txt` currently makes both **pkg-config** and **libjxl** hard build-time requirements:

```cmake
FIND_PACKAGE(PkgConfig REQUIRED)
pkg_check_modules_fullpath_libs(JPEGXL libjxl)
```

…and inside `pkg_check_modules_fullpath_libs`:

```cmake
pkg_check_modules(${PREFIX} REQUIRED IMPORTED_TARGET ${MODULE_NAME})
```

So the configure step fails (rather than continuing without JXL) on any machine where:

- pkg-config isn't installed — common on macOS/Windows where dependencies come from Homebrew, conan, or vcpkg rather than pkg-config; or
- pkg-config is installed but `libjxl.pc` isn't.

The macro already has a `# not found, support will be disabled` else-branch, but with `REQUIRED` on `pkg_check_modules` it's unreachable.

## Changes

- `FIND_PACKAGE(PkgConfig REQUIRED)` → `FIND_PACKAGE(PkgConfig QUIET)`, and the call to `pkg_check_modules_fullpath_libs(JPEGXL libjxl)` is only made when `PkgConfig_FOUND`.
- Inside `pkg_check_modules_fullpath_libs`, `REQUIRED` is dropped from `pkg_check_modules` so the macro's existing "not found, support will be disabled" path is reachable.

Two-line change to a single file. No source code changes.

## Behavior

| Environment | Before | After |
|---|---|---|
| pkg-config + libjxl present | builds with JXL | builds with JXL (unchanged) |
| pkg-config present, libjxl missing | configure fails | builds without JXL |
| pkg-config not installed | configure fails | builds without JXL |

## Context

Hit this writing a Conan recipe for OpenMVS targeting macOS Apple Silicon. Conan's hermetic builds intentionally don't expose system pkg-config to the project, and adding pkg-config + a libjxl pc file as a hard build dependency for an optional codec didn't seem right.

Happy to address review comments.


ShuangLiu1992:
@cdcseacave PTAL

cdcseacave:
@claude review

cdcseacave:
thank you

PR #1272, opened 2026-05-16
Allow disabling subdirectory on install

sylveon:
This option is useful if your vcpkg triplet is x64-windows, because it causes DLLs to be used for dependencies like boost, CDAL, and etc.

Because OpenMVS uses a subdirectory for its binaries, it means that apps cannot find their DLL dependencies, even if OpenMVS itself is statically linked. This adds a simple option to disable that, keeping existing behavior unless it is flipped off.


cdcseacave:
Thanks @sylveon — this fix has been incorporated into https://github.com/cdcseacave/openMVS/pull/1286

PR #1273, opened 2026-05-20
Fix build with current Boost and PoseLib

Rheinwalt:
This updates the build and SFM PoseLib integration for current dependency versions.

### Changes
  - Prefer modern Boost Python CMake targets when available, while keeping fallback behavior.
  - Adapt bearing pose estimation calls to the current PoseLib API behind a small compatibility wrapper.

### Tested

On Arch Linux with:

- Boost 1.91
- Python 3.14
- PoseLib master `012f52ab5183cb72b4f3cdca5d389ef556f0a107`


cdcseacave:
thank you for all the work, but I can not merge it, it is too intrusive and deviates from the coding style (ex replace GIT_LOG() with std::cout)

reduce this to build fix only

Rheinwalt:
Like this maybe?

cdcseacave:
why did you create the two poselib files?

Rheinwalt:
The two files are needed for me on arch linux in order to build against vanilla upstream PoseLib. My guess is you're using a custom/old poselib which exposes estimate_relative_pose_bearings / estimate_absolute_pose_bearings?

Rheinwalt:
**Rebased onto current develop, including the upstream PoseLib bearing API changes.**

The remaining changes in PairsMatcher.cpp and Resection.cpp are not conflict-resolution code. They route the bearing-vector pose calls through a small compatibility adapter so OpenMVS can build both with the PoseLib API expected by current develop and with vanilla upstream PoseLib master, which does not expose the same estimate_*_pose_bearings entry points.

cdcseacave:
Thank you @Rheinwalt for identifying the modern Boost.Python build issue and testing it against current dependencies. The build portion has been consolidated into #1294, where this PR and your contribution are explicitly credited. The replacement keeps Python/Boost.Python target discovery private to `pyOpenMVS` and preserves OpenMVS's pinned bearing-aware PoseLib API instead of carrying a compatibility implementation. I will leave this PR open until #1294 merges, then close it as superseded rather than merge the older commits.

cdcseacave:
Merged as part of #1294 (e2f38b4f) — thank you for tracking down the build breakage against current Boost.

The merged version links `Python3::Module` and keeps Boost.Python private to the `pyOpenMVS` target rather than leaking Python into the core libraries, so the versioned-component lookup no longer has to succeed for the rest of the build.

One addition from review: current Boost config packages often expose only the un-versioned `Boost::python` target, and accepting it silently can link a Boost.Python built against a different CPython than the interpreter `FindPython3` selected — which surfaces only as an undefined-symbol failure when importing the module. The fallback is kept, since often nothing else is available, but configure now warns that it cannot be verified against the interpreter.

Also folded in from the same area: the unreachable Python branch after a `REQUIRED` `find_package`, and the two Python library lists collapsed into one since they were only ever consumed together.

The CGAL border API and the deprecated `wstring_convert` warnings you raised are handled in the same merge — the CGAL version check is now resolved once at the include site via a namespace alias, and the wide-string conversion is a hand-rolled UTF-8 encoder that also handles surrogate pairs correctly.

PR #1275, opened 2026-05-28
dense: parallelize PatchMatch CUDA across workers (~30% wall time)

cdcseacave:
## Summary

Parallelizes the PatchMatch CUDA path in dense reconstruction. On a 3796-image / 4K dataset wall time drops ~33% (44m43s -> ~30m); on the 12-image garagedoor smoke test 2.1s -> 1.2s. Fused-point count delta <0.5% (parallel non-determinism).

The fix is a stack of five changes that compound:

- **Narrow the kernel-launch mutex** to cover only `RunCUDA()`, not all of `EstimateDepthMap` — host-side prep (image load, neighbor selection, depth-prior packing) runs lock-free.
- **Per-instance `cudaStream_t`** lazily created on first `EstimateDepthMap` call (the dense-recon worker thread differs from the constructor's), replacing the default stream for every `cudaMemcpy*Async` and kernel launch.
- **Chain `__constant__` writes via a `cudaEvent_t`** so cross-instance ordering on module-global constant memory is preserved without holding a host mutex across the kernel.
- **Per-worker `PatchMatchCUDA` pool** sized via the new `--patch-match-cuda-instances` CLI / `OPTDENSE::nPatchMatchCUDAInstances` config (default 4, clamped to `nMaxThreads`). Workers claim a slot via a thread-local index gated by an epoch counter so slot ownership resets cleanly between the depth-map and geometric-consistency phases. The dense `ThreadPool` and the `data.sem` capacity are raised to match.
- **Pinned host staging for H->D image and depth-prior uploads** above 1.5 MP. `cv::Mat` sources for `cudaMemcpy2DToArrayAsync` are pageable; the driver was synchronously bouncing through its internal pinned buffer. A per-instance `StagedUploadCvMat` allocates `cudaHostAlloc` slots on demand and reuses them across scales (safe because `RunCUDA` ends with `cudaStreamSynchronize`).

Quality and correctness are unchanged. Geometric-consistency phase uses the same pool size; spurious cached `.dmap` files from prior runs still need to be cleared between resolution-level changes (unrelated, longstanding).

## Test plan

- [x] Smoke test: 12-image garagedoor scene completes, point count matches baseline within parallel-non-determinism noise
- [x] Large scene: 3796-image scene completes ~30% faster, fused point count within 0.5% of baseline
- [x] `--patch-match-cuda-instances 1` reproduces single-threaded behavior (degenerate pool)
- [x] Geometric-consistency phase parallelism verified (same `data.nDenseWorkers` count)
- [ ] CI build on Linux + Windows

### DIFF
--- a/apps/DensifyPointCloud/DensifyPointCloud.cpp
+++ b/apps/DensifyPointCloud/DensifyPointCloud.cpp
@@ -137,6 +137,7 @@ bool Application::Initialize(size_t argc, LPCTSTR* argv)
 	unsigned nSubResolutionLevels;
 	unsigned nEstimationIters;
 	unsigned nEstimationGeometricIters;
+	unsigned nPatchMatchCUDAInstances;
 	unsigned nEstimateColors;
 	unsigned nEstimateNormals;
 	unsigned nFuseFilter;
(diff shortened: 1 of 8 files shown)


cdcseacave:
@claude review

PR #1276, opened 2026-05-28
Fix bad_alloc in verbose depth-map point-cloud export

Rheinwalt:
Fix a crash when running DensifyPointCloud with high-resolution depth maps and verbose logging enabled.

At `-v 3`, dense reconstruction exports per-depth-map debug PLY files. The preallocated PLY memory buffer size can overflow for large depth maps, producing a huge allocation request and aborting with `std::bad_alloc`.

This avoids the preallocated memory buffer for these debug point-cloud exports and uses the normal streamed PLY write path instead.

### DIFF
--- a/libs/MVS/DepthMap.cpp
+++ b/libs/MVS/DepthMap.cpp
@@ -1927,6 +1927,11 @@ bool MVS::ExportPointCloud(const String& fileName, const Image& imageData, const
 {
 	ASSERT(!depthMap.empty());
 	const Camera& P0 = imageData.camera;
+	size_t numPoints(0);
+	for (int i=0; i<depthMap.area(); ++i) {
+		if (depthMap[i] > 0)
+			++numPoints;
+	}
 	if (normalMap.empty()) {
 		// vertex definition
 		struct Vertex {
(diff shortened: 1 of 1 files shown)


cdcseacave:
thank you for spotting the bug, but PLY writing with buffer is much faster, instead of removing it can you fix the buffer size approximation pls?

Rheinwalt:
Maybe you're right, no buffer was the lazy way out. However, with many threads and large images this can be pretty memory intensive still.

cdcseacave:
Thank you @Rheinwalt for identifying the verbose depth-map PLY `bad_alloc` path. It has been consolidated into #1294, where this PR and your contribution are explicitly credited. The replacement preserves fast buffering for small exports, caps it at 64 MB, and streams larger outputs. I will leave this PR open until #1294 merges, then close it as superseded rather than merge the older commit.

cdcseacave:
Merged as part of #1294 (e2f38b4f) — thank you, the `bad_alloc` you reported was real and the sizing was unbounded.

The merged version caps the in-memory body at 64 MB and takes PLY's streaming path above that. Two things were added on top of your fix:

- The sizing helper was hoisted out of `DepthMap.cpp` into `PLY::ComputeMemBufferSize`, because the identical unbounded `area() * bytes-per-point` pattern also existed in `SemiGlobalMatcher::ExportPointCloud`, which could not have reused a file-local helper. Both exporters and `Scene::ExportLinesPLY` now share one rule.
- An edge case surfaced while reviewing it: a depth-map with no valid depths produces a zero-sized buffer, and zero is also `PLY::write`'s "stream directly to disk" sentinel. That path creates and truncates the destination before the export bails out on zero elements, leaving a stray 0-byte `.ply` and destroying any valid export already at that path. Both exporters now stop before touching disk when there is nothing to write.

The point count is also obtained with `cv::countNonZero` rather than a scalar pass over the map.

PR #1279, opened 2026-06-08
Add Metal compute backend for PatchMatch densification (Apple Silicon)

leNeo:
## Summary

Adds a **Metal compute backend** for PatchMatch dense densification, giving GPU-accelerated `DensifyPointCloud` on **Apple Silicon**, where CUDA is unavailable. It ports `PatchMatchCUDA` (ACMH) to Metal Shading Language and wires it into `SceneDensify` alongside the existing CUDA path.

Opening as a **draft** to gather feedback on the approach before finalizing (see follow-ups).

## What's included

- **`libs/MVS/PatchMatchMetal.metal`** — full MSL port of the 4 kernels (`InitializeScore`, `Black`/`RedPixelProcess`, `FilterPlanes`) and every device function (checkerboard propagation, Monte-Carlo view selection, 3-way perturbation refinement, geometric-consistency weight). Translation notes: `curand` → counter-based PCG; `__constant__` cameras/params → buffers; the `<bool GEOM>` template → a Metal function constant.
- **`libs/MVS/PatchMatchMetal.{h,mm}`** — pure-C++ PIMPL header (so plain C++ TUs can hold the class) + Objective-C++ host mirroring `PatchMatchCUDA.cpp`: multi-resolution loop, camera/image upload to `MTLTexture`, depth/normal/conf/views readback. Supports photometric **and** geometric-consistency passes.
- **`SceneDensify`** — `pmMetalPool` wired next to `pmCUDAPool` using the same per-worker slot mechanism; `ReinitMetalPoolForGeom` for the geom phase. `OPENMVS_DISABLE_METAL=1` forces the CPU path.
- **CMake** — `_USE_METAL` auto-enabled on Apple (`ENABLE_LANGUAGE(OBJCXX)`, Metal/Foundation frameworks). The MSL is embedded from the single-source `.metal` at configure time and compiled at runtime via `newLibraryWithSource`. On non-Apple platforms nothing changes.

## Validation (bundled 4-image test scene)

- **Geometric accuracy** — where Metal and the CPU path both produce a depth, they agree to **~0.10–0.11 % median** relative difference → camera/homography/triangulation are correct (incl. real rotations).
- **Performance** — **~5–6× faster** wall-time; CPU-user time drops from ~18 s to ~0.1 s (work genuinely on the GPU). Geometric-consistency passes run ~20× faster than the CPU path.
- **Completeness** — Metal raw depth-maps are *more* complete (≈98 % valid px vs ≈78 % for the CPU `DepthEstimator`). The fused-cloud point count is lower than the CPU path; this reflects an **algorithm difference** (a faithful PatchMatch**CUDA** port vs the separate CPU `DepthEstimator`), not a regression — the true reference is the CUDA backend, which I could not run (no NVIDIA GPU).

## Known limitations / follow-ups

- Buffers/textures are re-allocated per scale per call; caching across calls would cut overhead.
- MSL is compiled at runtime from embedded source; a build-time `.metallib` would be cleaner for distribution.
- Validated on a small scene only; large drone-scale benchmarks pending.
- Geometric-consistency is implemented and runs; cross-validation against the CUDA backend would be valuable.


cdcseacave:
@leNeo great contribution, this PR is very valuable; pls help with the last part in order to be able to merge it: full testing on various datasets and configurations and compare to the existing CUDA implementation (if in terms of speed will be dificult on different architectures, it will be very easy to compare the quality and stability); use TnT datasets for evaluating F1 score with both implmentations and present a full table with all the results (including time even if different PCs)

leNeo:
@cdcseacave thanks! I've added a fully reproducible benchmark harness under
[`eval/tnt/`](https://github.com/leNeo/openMVS-metal/blob/metal-patchmatch-backend/eval/tnt/README.md) and ran the first comparison.

### Methodology (one variable changed)

SfM is run once per scene → a single shared `scene.mvs`; every backend densifies
*that identical file* with *identical* fixed parameters, and the *same* GT + the
official [TanksAndTemples](https://github.com/isl-org/TanksAndTemples) toolbox
scores every result. So the F-score isolates the backend, not the configuration.

### Metal vs CPU — Tanks-and-Temples training set (F-score @ default τ)

Both backends ran on the **same machine** (Apple M4 Max, 48 GB), so here the
wall-times *are* directly comparable; the CPU path is the multithreaded
`DepthEstimator`, Metal is the new PatchMatch backend.

| Scene | Backend | F1 ↑ | Precision | Recall | Points | Wall (s) |
|---|---|---|---|---|---|---|
| Barn | metal | **0.409** | 0.378 | 0.446 | 6.87M | 64.7 |
| Barn | cpu | 0.371 | 0.326 | 0.430 | 7.87M | 481.0 |
| Ignatius | metal | **0.657** | 0.592 | 0.738 | 4.39M | 38.6 |
| Ignatius | cpu | 0.608 | 0.525 | 0.720 | 5.07M | 348.7 |
| Meetingroom | metal | **0.239** | 0.417 | 0.168 | 2.68M | 55.2 |
| Meetingroom | cpu | 0.215 | 0.331 | 0.159 | 3.40M | 470.3 |
| Truck | metal | **0.647** | 0.597 | 0.707 | 4.97M | 40.2 |
| Truck | cpu | 0.593 | 0.519 | 0.693 | 5.49M | 366.4 |
| **mean** | **metal** | **0.488** | | | | **49.7** |
| **mean** | **cpu** | 0.447 | | | | 416.6 |

**Takeaways**
- The Metal backend **matches or slightly exceeds** the CPU `DepthEstimator`
  F-score on every scene (mean **0.488 vs 0.447**), at **~8× faster** wall time
  on identical hardware.
- Parity holds even on the hard indoor scene (Meetingroom), where both backends
  have low absolute recall — Metal is still ahead.
- The CPU path produces more raw points but with lower precision; Metal's clouds
  are leaner and slightly more accurate after fusion.

This corroborates the earlier depth-map analysis in this PR (median depth
agreement ~0.1% where both are valid; Metal depth-maps actually *more* complete).

### CUDA column — call for help

I don't have an NVIDIA GPU, so I can't produce the CUDA PatchMatch numbers
locally. To keep it perfectly controlled, the harness publishes the shared
`scene.mvs` files and a one-shot
[`contribute_cuda.sh`](https://github.com/leNeo/openMVS-metal/blob/metal-patchmatch-backend/eval/tnt/contribute_cuda.sh): a volunteer densifies the
*same* `scene.mvs` with a CUDA build and sends back the fused PLY + timing — no
GT or toolbox needed on their side; I run the identical F1 evaluation here so the
result drops straight into the table. If you or anyone with a CUDA box can run it
on even one or two of these scenes, the comparison is complete.

Shared inputs (scene.mvs + undistorted images), one self-contained tarball per
scene — [**release: tnt-bench-inputs**](https://github.com/leNeo/openMVS-metal/releases/tag/tnt-bench-inputs):
[Barn](https://github.com/leNeo/openMVS-metal/releases/download/tnt-bench-inputs/Barn_mvs.tar.gz) ·
[Ignatius](https://github.com/leNeo/openMVS-metal/releases/download/tnt-bench-inputs/Ignatius_mvs.tar.gz) ·
[Meetingroom](https://github.com/leNeo/openMVS-metal/releases/download/tnt-bench-inputs/Meetingroom_mvs.tar.gz) ·
[Truck](https://github.com/leNeo/openMVS-metal/releases/download/tnt-bench-inputs/Truck_mvs.tar.gz)

> The shared `scene.mvs` bundles undistorted frames derived from the
> [Tanks and Temples](https://www.tanksandtemples.org/) training dataset
> (Knapitsch et al., *ACM ToG* 2017), redistributed for non-commercial
> research under [CC BY-NC-SA](https://creativecommons.org/licenses/by-nc-sa/3.0/).

cdcseacave:
thank you, I will run CUDA version and complete the results here

cdcseacave:
Ran the CUDA PatchMatch backend through the harness so the missing **CUDA column**
is now filled. To keep it a true one-variable comparison I densified the **exact
shared `scene.mvs` files** published in `tnt-bench-inputs` (same SfM the Metal/CPU
rows used), with the **identical** `eval/tnt` params
(`-R1 --min-resolution 640 --max-resolution 3200 --number-views 5 --fusion-mode 0`),
and scored every cloud with the same official TanksAndTemples toolbox at the same
per-scene τ. The CUDA trajectory was derived from the shared `scene.mvs`
(`InterfaceCOLMAP` export → `colmap_to_tnt_log.py`), entry counts match the
official `_COLMAP_SfM.log` exactly for all four scenes.

### Metal vs CPU vs CUDA — Tanks-and-Temples training set (F-score @ official τ)

Same shared `scene.mvs`, same params, same GT/toolbox — only the PatchMatch
backend differs. Metal/CPU wall-times are from the PR (Apple M4 Max); CUDA
wall-times are on an NVIDIA GPU (i7-13700KF host) and are **not**
cross-architecture comparable.

| Scene | Backend | F1 ↑ | Precision | Recall | Points | Wall (s) |
|---|---|---|---|---|---|---|
| Barn | metal | 0.409 | 0.378 | 0.446 | 6.87M | 64.7 |
| Barn | cpu | 0.371 | 0.326 | 0.430 | 7.87M | 481.0 |
| Barn | **cuda** | **0.424** | 0.387 | 0.471 | 7.16M | 253.8 |
| Ignatius | metal | 0.657 | 0.592 | 0.738 | 4.39M | 38.6 |
| Ignatius | cpu | 0.608 | 0.525 | 0.720 | 5.07M | 348.7 |
| Ignatius | **cuda** | **0.668** | 0.602 | 0.750 | 4.55M | 71.6 |
| Meetingroom | metal | 0.239 | 0.417 | 0.168 | 2.68M | 55.2 |
| Meetingroom | cpu | 0.215 | 0.331 | 0.159 | 3.40M | 470.3 |
| Meetingroom | **cuda** | **0.253** | 0.403 | 0.184 | 3.11M | 105.9 |
| Truck | metal | 0.647 | 0.597 | 0.707 | 4.97M | 40.2 |
| Truck | cpu | 0.593 | 0.519 | 0.693 | 5.49M | 366.4 |
| Truck | **cuda** | **0.649** | 0.600 | 0.708 | 5.09M | 69.9 |
| **mean** | metal | 0.488 | | | | |
| **mean** | cpu | 0.447 | | | | |
| **mean** | **cuda** | **0.499** | | | | |

**Takeaways**
- **CUDA and Metal land on top of each other** — within ~0.01–0.015 F1 on every
  scene (mean **0.499 vs 0.488**), and CUDA is a hair ahead on all four. This is
  the direct confirmation the PR was missing: the Metal port reproduces the CUDA
  reference's quality, it isn't just beating the separate CPU `DepthEstimator`.
- Both GPU backends clearly beat the CPU path (mean 0.447), consistent with the
  Metal-vs-CPU result already in the PR.
- Per-backend precision/recall split is also nearly identical between CUDA and
  Metal (e.g. Truck 0.600/0.708 cuda vs 0.597/0.707 metal), so the agreement is
  not a lucky F1 cancellation — the clouds themselves match.

<details>
<summary>Sanity check on an independent SfM (local Metashape `scene.mvs`)</summary>

CUDA, same params/toolbox, but a different SfM front-end (Metashape instead of the
PR's COLMAP `scene.mvs`) — **not** comparable to the table above, just a second
independent reconstruction to show the backend is stable across inputs:

| Scene | F1 ↑ | Precision | Recall | Points | Wall (s) | τ |
|---|---|---|---|---|---|---|
| Barn | 0.559 | 0.543 | 0.575 | 7.43M | 108.9 | 0.010 |
| Ignatius | 0.706 | 0.668 | 0.749 | 4.43M | 76.0 | 0.003 |
| Meetingroom | 0.345 | 0.523 | 0.257 | 3.83M | 158.1 | 0.010 |
| Truck | 0.695 | 0.672 | 0.720 | 5.26M | 79.0 | 0.005 |

The absolute F1 is higher here purely because the Metashape SfM is a stronger
starting point — it underlines that the SfM front-end, not the densify backend,
drives the absolute numbers, which is exactly why the table above pins `scene.mvs`.
</details>

cdcseacave:
@leNeo I run the CUDA version with your data and in general the F1 score is 1-2% higher, can you pls finetune a bit the metal implementation?

leNeo:
Found and fixed it — the geometric-consistency pointer was on the money.

**Root cause:** `GeometricConsistencyWeight` collapsed two cases the CUDA reference keeps distinct — a neighbor with *no* depth-map (`if (depthImage == NULL) return 0.f` → ignore the term) vs. a depth-map whose projected pixel is invalid (→ `maxDist` penalty). Metal binds a 1×1 zero dummy texture for absent depth-maps, so the `trgDepth == 0` branch was charging the full penalty for *missing* maps too — making Metal systematically more conservative than CUDA (fewer fused points, lower recall).

**Fix:** detect the dummy by its size and return 0, mirroring CUDA's NULL check — one line, no new bindings (the dummy is always 1×1, real depth-maps never are).

**Result** (TnT training, shared `scene.mvs`, identical params): mean F1 **0.488 → 0.494** (CUDA 0.499), with recall up on all four scenes. Biggest gain on Meetingroom **0.239 → 0.252** (CUDA 0.253) — the indoor scene where neighbors most often lack a depth-map, exactly where the over-penalization hurt most.

| Scene | metal (before) | metal (fixed) | cuda |
|---|---|---|---|
| Barn | 0.409 | 0.415 | 0.424 |
| Ignatius | 0.657 | 0.663 | 0.668 |
| Meetingroom | 0.239 | 0.252 | 0.253 |
| Truck | 0.647 | 0.648 | 0.649 |
| mean | 0.488 | **0.494** | 0.499 |

Pushed as `f819795`.

leNeo:
Heads-up on a concurrency bug I hit while stress-testing the Metal backend with higher iteration counts — **it's in the shared worker orchestration, not Metal-specific, and likely affects the CUDA path too.** Pushed a fix in `53285e7`.

**Symptom:** with `--iters` or `--geometric-iters` above the defaults, `DenseReconstruction` aborted non-deterministically — `exit(1)` with no crash report (so a clean `return false`, not a GPU fault). Single-threaded and under lldb it always succeeded → a race.

**Root cause:** `DenseReconstructionEstimate` ends each worker phase by broadcasting one `EVT_CLOSE` per sibling. With `nDenseWorkers > 2` (the Metal **and CUDA** pools default to 4) the shutdown handshake can leave orphaned `EVT_CLOSE` sentinels queued when the threads join, so `if (!data.events.IsEmpty()) return false;` fires spuriously. The check runs once after the photometric phase **and once per geometric iteration**, so more geo-iters multiplies the race window — which is why it only showed up when pushing iteration counts up.

Since the CUDA pool is also 4 workers, I suspect this can bite the CUDA path on your Linux CI under the same conditions (it just hasn't been exercised with raised iteration counts) — worth a look on your side where you can reproduce deterministically.

**Fix:** stop overloading the event queue as a status signal. Added an explicit `nFailures` flag for genuine worker errors, check that instead, and `Clear()` the queue between phases so leftover sentinels don't carry into the next one. Stress test (`--geometric-iters 4`, 4 Metal workers): 6/6 clean runs, identical 4567870 fused points (was failing ~half the time); default-config F1 unchanged.

cdcseacave:
Followed up on the worker-orchestration race from `53285e7` — good catch, it's a
real latent bug in the shared `DenseReconstructionEstimate` shutdown, not
Metal-specific. Two things from my side: a reproduction attempt on CUDA, and a
root-cause fix pushed to the branch (`00ace26`).

### Reproduction on CUDA (RTX, i7-13700KF)

I couldn't trigger it on the CUDA path, even at high exposure:

| Build | Config | Result |
|---|---|---|
| pre-fix (`f819795`) | Ignatius, `--geometric-iters 6` (7 phases/run), 16 CUDA workers, 8 runs | **8/8 OK, 0 aborts** |
| fixed (`00ace26`) | same | **8/8 OK, 0 aborts**, identical point counts |

That lines up with the mechanism. The orphan condition needs **≥2 workers reaching
the end-of-work branch in the same window** (then the broadcast leaves
`nDenseWorkers·(K-1)` stray `EVT_CLOSE`). On the CUDA path the per-device kernel
serializer staggers when workers finish their last image, so K stays 1 almost
always — which is most likely why it never showed up in CI here, while it
reproduces readily on Metal where the workers finish in lockstep. So: real shared
bug, Metal-*exposed*, latent-but-rare on CUDA. Worth fixing regardless.

### Root-cause fix (`00ace26`)

Rather than treat a non-empty queue as "not a failure" (the `nFailures` flag in
`53285e7` was never actually incremented, so the guard became a no-op; I'd dropped
it to just `events.Clear()` in `c71d10e`), this fixes the shutdown so the queue is
genuinely empty after a clean phase — which lets `if (!events.IsEmpty()) return
false;` go back to being a *reliable* failure signal (same as the filter phase):

- a per-phase atomic latch `nClosing`: only the **first** worker at the end-branch
  (`safeInc → 1`) enqueues **exactly one `EVT_CLOSE` per worker**, itself included;
- every worker — including the ones that hit the end-branch, which now `break` to
  loop back — exits by consuming **exactly one** close.

Counts stay balanced → no orphaned sentinels → empty queue on success. Removed both
`events.Clear()` calls and restored the `!events.IsEmpty()` checks in the
photometric and per-geometric-iteration joins.

Validated: 6/6 + 8/8 clean runs at `--geometric-iters 4`/`6` with 12–16 CUDA
workers, fused point counts identical to pre-fix (no behavioral change on the happy
path). If you can re-run your Metal stress test (`--geometric-iters 4`, 4 workers)
against `00ace26`, that would confirm it closes the window on the path that
actually triggers it.

cdcseacave:
@leNeo pls retest the final version of the code when you have time and confirm it is working, and if there is anything else you want to add, we are clear to squash onto develop

leNeo:
Confirmed on Metal — `00ace26` closes it. Rebuilt the branch on Apple Silicon (M4 Max) and stress-tested the path that actually triggers the orphan (Metal, 4 workers, lockstep finish):

| Config | Phases/run | Result |
|---|---|---|
| `--geometric-iters 4` | 5 | **6/6 clean** |
| `--geometric-iters 6` | 7 | **4/4 clean** |

10/10, zero aborts. Every run is byte-deterministic (geom-iters 4 → 4567870 pts every time, geom-iters 6 → 4588250), and the geom-iters-4 count is **identical to pre-fix** — so the `nClosing` latch has no behavioral effect on the happy path, it just balances the EVT_CLOSE count.

Before the fix this exact config aborted ~50% of runs, so the window is closed on the path where it reproduces. Your mechanism explanation matches what I see: on Metal the workers finish their last image in lockstep, so ≥2 routinely hit the end-branch in the same window (K≥2) — exactly the orphan condition — whereas the CUDA serializer staggers them to K=1. Nice root-cause fix; restoring `!events.IsEmpty()` as a real failure signal is much cleaner than my drain workaround.

cdcseacave:
Pushed a few cleanups on top of the branch (`f6a4dd0..e077fde`):

### 1. Unified GPU backend selection under `--gpu-device` (`75ffb42`)
Rather than toggling Metal with the `OPENMVS_DISABLE_METAL` environment variable, both backends are now selected the same way — through a single `--gpu-device` flag (CUDA on Windows/Linux, Metal on Apple). It replaces the old `--cuda-device`:
- `-1` → best/default GPU, `>=0` → comma-separated device IDs, `-2`/`cpu`/`none`/empty → CPU.
- The device string and `isCpuRequested()` were moved out of the `#ifdef _USE_CUDA` block in `UtilCUDA` so they're available to every build (Metal and CPU-only included); `UtilCUDA.cpp` is always compiled, so this links everywhere.
- The Metal pool gate now uses `isCpuRequested(desiredDeviceIDs)` exactly like CUDA. Since the two gates became identical, the CUDA and Metal pool-allocation blocks in `SceneDensify` are merged into one (only the `Allocate*Pool` call differs), and both now log the active backend at verbose level.
- `--cuda-device` → `--gpu-device` across `DensifyPointCloud`, `RefineMesh`, `ExtractKeyframes` and the Viewer UI; the densify flag is guarded by `_USE_CUDA || _USE_METAL`.
- `forceCPU` in the pipeline test now also disables the Metal path.
- The T&T eval scripts/docs (`densify.sh`, `contribute_cuda.sh`, `README.md`) use `--gpu-device` instead of the env var.

> Note: this drops `--cuda-device` — existing scripts/configs that used it need updating to `--gpu-device` (same syntax). Happy to add a hidden deprecated alias if you'd prefer a softer transition.

### 2. `CreateStructure --undistort-extension` (`36c7f89`)
The undistorted images exported by `--export-mvs` were always written as `.jxl` (the `ExportMVSConfig::extension` default), with no way to change it from the CLI. Added `--undistort-extension` (default `.jxl`) to pick the format, e.g.:
```
CreateStructure -s imgs -o scene.sfm --export-mvs scene.mvs --undistort-extension .jpg
```

### 3. License headers on the Metal files (`e077fde`)
Added the standard SEACAVE AGPL header to `PatchMatchMetal.h/.mm/.metal` to match the rest of the tree, with the contributor credited in each file's description.

**Validation (Windows, CUDA build, RelWithDebInfo):** clean build, `--gpu-device`/`--undistort-extension` present in the binaries (no stale `cuda-device`), and `ctest -R MVSPipelineTest` passes. The Metal path is logic-reviewed only (can't compile here) — a quick rebuild on Apple to confirm `--gpu-device cpu`/`-1` behave as expected would be appreciated.

PR #1282, opened 2026-06-15
TextureMesh: export vertex-colored PLY meshes

Rheinwalt:
This adds a `TextureMesh --vertex-colors` mode for exporting a PLY mesh with per-vertex RGB colors.

This is related to the older vertex-coloring proposal in #424, but takes a smaller export-only approach:

 - does not add persistent vertex colors to `Mesh`
 - does not change `Mesh::SavePLY`
 - does not generate a texture atlas
 - samples source image colors directly at mesh vertices after face best-view assignment
 - averages multiple contributing face/view samples per vertex, weighted by face area, instead of letting the last sampled face overwrite the vertex color

The tradeoff is that this does not currently use seam leveling/exposure correction from the full texture-atlas pipeline. The intended use is a lightweight colored PLY export when a full texture atlas is unnecessary or too expensive.

The PLY writer outputs standard `red`, `green`, `blue` uchar vertex properties in RGB order.

### DIFF
--- a/apps/TextureMesh/TextureMesh.cpp
+++ b/apps/TextureMesh/TextureMesh.cpp
@@ -69,6 +69,7 @@ int nProcessPriority;
 unsigned nMaxThreads;
 int nMaxTextureSize;
 bool bExportTextureLossless;
+bool bVertexColors;
 String strExportType;
 String strConfigFileName;
 boost::program_options::variables_map vm;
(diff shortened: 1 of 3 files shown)


cdcseacave:
Thank you @Rheinwalt for proposing the lightweight vertex-colored PLY export and the area-weighted sampling approach. It has been consolidated into #1294, where this PR and your contribution are explicitly credited. The replacement reuses the existing PLY serializer, adds mask/resource handling, and tests ASCII/binary geometry and RGB channel order. I will leave this PR open until #1294 merges, then close it as superseded rather than merge the older commits.

cdcseacave:
Merged as part of #1294 (e2f38b4f) — thank you, this is a genuinely useful capability and the merged work is built directly on your design.

The feature was reworked during review so the colors are a first-class mesh attribute rather than a side channel. `Mesh::vertexColors` now sits beside `vertexNormals`, which means:

- colors round-trip through PLY instead of being export-only (previously the loader was capped at six properties, so a colored PLY silently lost its colors on reload — including re-running `TextureMesh` on its own output)
- they are stored in the `.mvs` archive, which required bumping the project stream version
- `SavePLY` no longer needs a `pVertexColors` parameter, and the app falls through to the shared save path instead of returning early from `main`
- the arrays that rebuild the vertex set (`RemoveVertices`, `RemoveDuplicatedVertices`, the CGAL round-trip in `Clean`) keep colors in sync or drop them, rather than leaving them stale

Two behavioral changes worth noting: sampling is grouped by view so each image and its mask are released once consumed (the earlier shape held every image plus a full-resolution mask until the function returned, which is many GB on a large scene), and the ignore mask is tested across the full bilinear footprint of each sample rather than a single pixel.

Coverage was added for the ASCII and binary PLY round-trip and for the coloring stage in the MVS pipeline test.

PR #1284, opened 2026-07-05
PatchMatchCUDA: fix out of bounds read.

sagielevy:
costArray[posId] is being sampled outside of the bounds of the image when sampling close to the edges. This is not only an out-of-bound read, but causes the behavior of the entire pipeline to produce non-deterministic results.


cdcseacave:
Thanks @sagielevy — this fix has been incorporated into #1286, where the per-position multi-view score aggregation now honors the `valid[]` flag so `costArray` is no longer sampled out of bounds near image edges (fixing both the OOB read and the resulting non-determinism). Closing in favor of #1286.

PR #1285, opened 2026-07-08
OBJ: Remove incorrect transparency on material

csparker247:
The OBJ MTL writer defaults to writing a fully _transparent_ material rather than an opaque one:
```
newmtl material_00
Ka 1.000000 1.000000 1.000000
Kd 1 1 1
Ks 0.000000 0.000000 0.000000
Tr 1.000000  # <- wrong
illum 1
Ns 1.000000
map_Kd
```
The spec default when Tr is not specified is `0.`, so this PR just removes this declaration entirely.

I ran into this with a downstream post-processor converting OBJs to GLBs and the post-processor made the materials transparent.


cdcseacave:
Thanks @csparker247 — this fix has been incorporated into #1286, where the incorrect `Tr 1.000000` (fully transparent) line is removed from the OBJ MTL writer so the spec-default opaque material is used. Closing in favor of #1286.

PR #1288, opened 2026-08-08
texture: honor ignore-mask-label instead of uninitialized OPTDENSE global

alesdev88:
## Problem

`TextureMesh --ignore-mask-label N` (N >= 0) produces an **inverted** result: the atlas is textured exclusively from the regions the mask asked to ignore, and every unmasked region is discarded.

## Root cause

`MeshTexture::ListCameraFaces` gates the mask import on its `nIgnoreMaskLabel` parameter (the CLI value), but then passes `OPTDENSE::nIgnoreMaskLabel` to `DepthEstimator::ImportIgnoreMask`:

```cpp
if (nIgnoreMaskLabel >= 0) {
    BitMatrix bmask;
    DepthEstimator::ImportIgnoreMask(imageData, fullSize, (uint8_t)OPTDENSE::nIgnoreMaskLabel, bmask, &rasterer.mask);
}
```

The TextureMesh app never initializes the OPTDENSE config, so the global is zero-initialized. `ImportIgnoreMask` then computes the validity mask as `(mask != 0)`: background pixels (0) are invalidated, and only pixels carrying the ignore label survive — an exact inversion of the intended behavior. DensifyPointCloud is unaffected because its app writes the OPTDENSE config before use.

## Repro

Per-image `.mask.png` files (person = 255, background = 0) on a 360-view indoor capture, `--ignore-mask-label 255`: the produced 4096 atlas contained **only the masked-out person**. With this one-argument fix, the same scene textures correctly with the person fully excluded (verified on macOS arm64, develop @ 69b25a3, OpenCV 4.14).

## Fix

Pass the function's own `nIgnoreMaskLabel` parameter through.

### DIFF
--- a/libs/MVS/SceneTexture.cpp
+++ b/libs/MVS/SceneTexture.cpp
@@ -552,7 +552,10 @@ bool MeshTexture::ListCameraFaces(FaceDataViewArr& facesDatas, float fOutlierThr
 		if (nIgnoreMaskLabel >= 0) {
 			// import mask
 			BitMatrix bmask;
-			DepthEstimator::ImportIgnoreMask(imageData, fullSize, (uint8_t)OPTDENSE::nIgnoreMaskLabel, bmask, &rasterer.mask);
+			// use the label passed by the caller: the OPTDENSE config is not
+			// initialized in the TextureMesh app, so its global would be 0 here,
+			// inverting the mask (only label-marked pixels would remain valid)
+			DepthEstimator::ImportIgnoreMask(imageData, fullSize, (uint8_t)nIgnoreMaskLabel, bmask, &rasterer.mask);
 		} else if (nIgnoreMaskLabel == -1) {
 			// creating mask to discard invalid regions created during image radial undistortion
 			rasterer.mask = DetectInvalidImageRegions(imageData.image);


cdcseacave:
Thank you @alesdev88 for the precise mask-label diagnosis and reproduction. This fix has been consolidated into #1294, where this PR and your contribution are explicitly credited. The replacement keeps the parameter-routing correction and also enforces the mask at vertex-color sample footprints. I will leave this PR open until #1294 merges, then close it as superseded rather than merge the older patch.

cdcseacave:
Merged as part of #1294 (e2f38b4f) — thank you for finding this.

The bug was exactly as you described: `MeshTexture::ListCameraFaces` gated on its own `nIgnoreMaskLabel` parameter but then passed `OPTDENSE::nIgnoreMaskLabel` to `ImportIgnoreMask`. Since `TextureMesh` never calls `OPTDENSE::init()`, that global holds whatever it was statically initialized to rather than the value behind `--ignore-mask-label`, so the option was effectively ignored in the texturing path.

Beyond the one-line fix, review pulled the mask policy (`>= 0` imports the stored mask, `-1` estimates the undistortion-invalid regions, `-2` disables) into a single `ComputeValidityMask` helper. Both face selection and the new vertex-color sampling path now call it, so the two cannot drift apart the way a second copy would have.

PR #1289, opened 2026-08-08
IO: resolve shared libraries in pkg_check_modules_fullpath_libs

alesdev88:
## Problem

Building `libs/IO` on macOS with Homebrew dependencies fails to link:

```
ld: library 'jxl' not found
```

## Root cause

`pkg_check_modules_fullpath_libs` converts pkg-config results into full-path libraries, but on non-Windows it only probes for **static** archives (`lib<name>.a`). Homebrew's `jpeg-xl` (like many of its packages) ships shared-only, so the probe fails and the macro silently falls back to the bare `-ljxl` with no accompanying link directory — the linker then can't resolve it. (The vcpkg CI builds never hit this because vcpkg produces `.a` archives.)

## Fix

Probe the platform shared-library extensions (`.dylib`, `.so`) after `.a` before falling back to the bare name. Verified on macOS arm64 (develop @ 69b25a3, Homebrew jpeg-xl 0.12): `libIO` and all app targets link, and `pkgcfg` full paths appear in the link lines.


cdcseacave:
Thank you @alesdev88 for diagnosing and reproducing the shared-library resolution failure. This proposal has been consolidated into #1294, where this PR and your contribution are explicitly credited. The replacement delegates path resolution to CMake's maintained `FindPkgConfig` result and includes the broader review/validation. I will leave this PR open until #1294 merges, then close it as superseded rather than merge the older patch.

cdcseacave:
Merged as part of #1294 (e2f38b4f) — thank you, this was a real bug and your diagnosis was correct.

The previous hand-rolled loop searched `${PREFIX}_LIBRARY_DIRS` for a `lib<name>.a` and otherwise fell back to the bare `-l` name, so libraries that did not match that shape were never resolved to a full path. The merged version takes your approach of resolving through CMake's maintained `${PREFIX}_LINK_LIBRARIES`.

One adjustment came out of review, which may be useful to know: passing `_LINK_LIBRARIES` straight through also drops the old preference for static archives, so on a host carrying both `.a` and `.so` the binaries silently switch to shared linkage and stop running where those libraries are absent. The merged code keeps your resolution and re-applies the static preference only for libraries inside the module's own `-L` directories.

That scoping matters more than it looks: applying the preference to every resolved entry rewrites `-lm` to the toolchain's `libm.a`, and glibc does not build those archives PIC, so `libIO.so` fails to link with `relocation R_X86_64_PC32 against undefined symbol '_dl_x86_cpu_features'`. We hit exactly that on Ubuntu CI before restricting it.

